PR #108, opened 2021-04-30
Split tests with long execution times into separate suite

dorranh:
Splits tests into a separate suite for the main set of tests we want to run on CI and longer-running tests which need to be run out-of-band for practical reasons. The separate suite is configured to run nightly in a new GitHub Actions workflow. I had to move the test "libs" (testCommon and testArbitary) into an actual lib to support using them in both test suites. `testAvl.ml` and `testLiquidationAuction.ml` also take a while to run, but I'm not sure if we want to stop checking these in the main CI pipeline.

Also fixed a small issue in the Makefile which caused the wrong shell to be called on the script for checking error codes if executed outside of a nix shell.

Happy to adjust this if another setup makes more sense - just let me know!


purcell:
I think we'll skip this PR in its current form, since execution time on CI now isn't a big issue. I'd still like it if the various property tests didn't take so long locally, so perhaps making the number of cases switchable using an env var is the way to go there, then crank it up in CI.

dorranh:
@purcell, thanks that sounds good to me.

PR #127, opened 2021-05-17
Add a sample e2e test using pytezos

utdemir:
This PR adds a sample test case using Pytezos, and adds the necessary Nix boilerplate to bring the Python stuff to the environment:

It introduces two top-level directories:

* `client/`: This is a Python library exporting a functions to deploy checker and other contracts. I think we should evolve this project to also be a command line application which can deploy checker, and possibly interact with a checker instance.
* `e2e/`: Includes the initial e2e test. I think we should keep these as simple as possible (without having too many assertions as we do on tests on the ocaml side), but it is worth checking each entrypoint at least once.

It introduces some dependencies:

* We use `poetry` to manage Python dependencies (mainly `pytezos` and `docker` libraries), and use a single lock file at the root of the project. These are made available on the Nix environment via `poetry2nix`. So, within `nix-shell`, we should just be able to use regular `python` command like `python e2e/main.py`.
* We depend on `ctez` repository as a submodule, because the tests needs to deploy it in order to use checkerL
* Running the end to end tests requires `ligo` and `tezos-client` binaries to be in scope, because of the `ctez` repository mentioned above which requires those in order to compile and deploy ctez. They are already in our Nix environment, so it shouldn't be too much of a problem for Linux users (sorry @purcell).

There're some pieces missing, but I would like to merge this and we should address them later on:

* CI doesn't run the test. 
* Testing other endpoints on e2e tests
* Having a cli
* Upstream the poetry2nix overrides

cc'ing @0xd34df00d and @dorranh


nix/python.nix
```diff
@@ -0,0 +1,53 @@
...
+    fastecdsa = super.fastecdsa.overridePythonAttrs (old: {
+      buildInputs = old.buildInputs ++ [ pkgs.gmp.dev ];
+    });
+    pendulum = super.pendulum.override {
+      preferWheel = true;
+    };
+    pytezos = super.pytezos.override (old: {
+      buildInputs = (old.buildInputs or [ ]) ++ [ pkgs.libsodium ];
+    });
+    # TODO: Upstream these overrides where possible.
```
dorranh, line 40: Ouch, that required quite a few overrides. Nice work :slightly_smiling_face:

e2e/main.py
```diff
@@ -0,0 +1,82 @@
...
+PROJECT_ROOT = os.path.join(os.path.dirname(__file__), "../")
+
+sys.path.append(os.path.join(PROJECT_ROOT, "client"))
+from checker import *
+
+class SandboxedTestCase(unittest.TestCase):
+    def setUp(self):
+       self.docker_client = docker.from_env()
+
+       port = portpicker.pick_unused_port()
```
dorranh, line 23: TIL about this

e2e/main.py
```diff
@@ -0,0 +1,82 @@
...
+
+         ports = { "20000/tcp": port },
+         name="checker-e2e-container-{}".format(port),
+
+         detach=True,
+         remove=True,
+       )
+
+       # wait some time for the node to start
+       time.sleep(10)
```
dorranh, line 37: I think the docker API exposes a status field which we might be able to use here in the future https://docs.docker.com/engine/api/v1.41/#operation/ContainerInspect

dorranh (commented):
Nice! In general LGTM, I waited to merge since I haven't been able to get the nix shell to build. Running into the following error when fetching MarkupSafe:

> Rewriting #!/nix/store/q6gfck5czr67090pwm53xrdyhpg6bx67-python3-3.8.9/bin/python3.8 to #!/nix/store/q6gfck5czr67090pwm53xrdyhpg6bx67-python3-3.8.9
  wrapping `/nix/store/r27p6x2vy9qv7yahmgv4cna2wwgc905v-python3.8-base58-1.0.3/bin/base58'...
  Executing pythonRemoveTestsDir
  Finished executing pythonRemoveTestsDir
  pythonCatchConflictsPhase
  pythonRemoveBinBytecodePhase
  pythonImportsCheckPhase
  Executing pythonImportsCheckPhase
  building '/nix/store/8lixv4f2h6y2sdxbab6nyp94ww9sj33j-MarkupSafe-2.0.0.tar.gz.drv'...
  trying https://files.pythonhosted.org/packages/source/m/MarkupSafe/MarkupSafe-2.0.0.tar.gz
    % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                   Dload  Upload   Total   Spent    Left  Speed
    0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0
  curl: (22) The requested URL returned error: 404 
  error: cannot download MarkupSafe-2.0.0.tar.gz from any mirror
  builder for '/nix/store/8lixv4f2h6y2sdxbab6nyp94ww9sj33j-MarkupSafe-2.0.0.tar.gz.drv' failed with exit code 1

I find this a bit weird since you can still access the detailed url from pypi is as follows: https://files.pythonhosted.org/packages/67/6a/5b3ed5c122e20c33d2562df06faf895a6b91b0a6b96a4626440ffe1d5c8e/MarkupSafe-2.0.0.tar.gz

e2e/main.py
utdemir: That'd be nice. However, I looked at the container and I don't think it exposes that info (`HEALTCHECK`, I think it is called).

However, I just looked at `pytezos` sources, and they do something like this:

```haskell
    def _wait_for_connection(cls) -> None:
        client = cls.get_client()
        while True:
            try:
                client.shell.node.get("/version/")
                break
            except requests.exceptions.ConnectionError:
                sleep(0.1)
```

I think it is a good idea, I'll do that.

e2e/main.py
utdemir: Yeah, it is handy. It's still not great since it doesn't guarantee the port staying empty until we bind to it later on, but should be good enough here.

nix/python.nix
purcell: I submitted some (but not all) of these overrides upstream: https://github.com/nix-community/poetry2nix/pull/305

Not sure what the rationale is for the msgpack and pendulum overrides, happy to add those too if they're generally useful.

utdemir:
> LGTM, I waited to merge since I haven't been able to get the nix shell to build. Running into the following error when fetching MarkupSafe:

Yeah, I got that after doing a garbage collection too. Lucky for us, there is an upstream PR already: https://github.com/nix-community/poetry2nix/pull/304  . But until that gets merged and ends up in nixpkgs; I'll hack it.

utdemir:
I fixed the nix issue, so I am merging this.

dorranh:
> Lucky for us, there is an upstream PR already

Oh nice, I had checked for issues on poetry2nix but had missed that PR. Thanks!

PR #128, opened 2021-05-18
Use cachix on CI

utdemir:
Since checker is now open source, we can easily use Cachix.

This PR also removes the `isCI` parameter that we used to have to avoid compiling tezos-client on CI, which now will come from the cache.


purcell:
Yay, thanks

PR #129, opened 2021-05-18
Minor fixes in the spec

gkaracha:
@purcell here are some small edits/feedback based on changes we made w.r.t. the original doc. There are two things that I was not sure how to fix:

1. The `Target` section in `design.rst` still is written with the understanding that the cfmm gives us tez/kit, but it actually gives us ctez/kit. Thing is, currently we _assume_ ctez/tez = 1 (which is not necessarily what we want to stick with) so technically the numbers there are still correct, but I guess that's subject to change(?). Maybe we should put a TODO there for when that part of the code is finalized? Not sure.
2. 
   > _"and saturates when the imbalance hits 20%"_

   This is no longer true, because the calculation changed in #124. Not sure what is the current percentage for saturation, but I'll tinker a little to figure it out, but I am sure it is not as pretty a number anymore. I wonder if we need this level of specificity here, since the parameter could change between checker instances. Might be a better idea to directly paste the formula from #124?

Also:

> _"ensures the stability of the target"_

"ensures" is probably a rather strong word here.

> _"tez are sometimes forfeited to it (TODO: still true?)"_

Yeah, this was probably never true (it certainly isn't true now) so I'd delete it altogether I think.

### DIFF
--- a/docs/spec/design.rst
+++ b/docs/spec/design.rst
@@ -41,7 +41,7 @@ to achieve a certain degree of steadiness with respect to a **target**,
 which is expressed in terms of a **quantity** and an **index**.
 
 The word “kit” is chosen because it’s short, simple to pronounce, and
-means a baby fox (which seem appropriate for a smart currency).
+means a baby fox (which seems appropriate for a smart currency).
 
 The **quantity**, :math:`q_{t_i}`, expressed in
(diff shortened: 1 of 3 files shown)


purcell:
Awesome, thanks George.

Issue #123, opened 2021-05-12
Make a command line client for deploying checker


utdemir:
Compiling and deploying checker is not trivial; since it requires a set of scripts to 

1. Transpilate ocaml to ligo
2. Separately compile the main contract and the lazy entrypoints
3. Deploy the main contract, then the lazy entrypoints
4. Seal the contract passing the necessary addresses for contracts it depends on

Steps 1 and 2 is required only when developing checker, so we can just share the resulting artifacts for deployment.

This issue is to create a script to do the step 3 and 4.

Currently we have a shell script to do the deployment; but I think using `pytezos` here will have some benefits on testing front, because then we can also write some end to end tests on Python, which is more plesant to write tests in than a shell script.

dorranh:
@utdemir, do you think that #127 + #130 cover this issue? Wanted to double check before closing.

utdemir:
Yes, definitely. Thanks!

PR #130, opened 2021-05-18
Add CLI to client library

dorranh:
Builds on #127 by adding a CLI for the Python client library for doing things like starting / stopping the sandbox and deploying contracts. This can be run in the nix-shell using the checker command, e.g. `checker --help`. Also updated the local deployment instructions in the README to use this new CLI.

Addresses #123


README.md
```diff
@@ -55,39 +55,18 @@ Generate the LIGO and Michelson code:
...
-$ tezos-client import secret key bob unencrypted:edsk3RFfvaFaxbHx8BMtEW1rKQcPtDML3LXjNqMNLCzC3wLC1bWbAt --force
-```
-
-And finally, deploy the contract (and packed entrypoints) to the sandbox:
+> Note: If no port is specified, the client will attempt to select a default one. To view the port number for
+> use with tezos-client, etc. you can use: `checker show-config`.
 
+And finally, deploy checker itself:
 ```console
-$ ./scripts/deploy-contract.sh
```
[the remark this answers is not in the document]
utdemir: And since we're at it, we can also delete the `run-sandbox.sh` too.

client/checker_client/cli.py
```diff
@@ -0,0 +1,229 @@
...
+    checker_address: str = ""
+
+    @staticmethod
+    def load(file: Path):
+        with file.open("r") as f:
+            return ConfigSchema().loads(f.read())
+
+    def dump(self):
+        with open(self.path, "w") as f:
+            f.write(ConfigSchema().dumps(self))
```
utdemir, line 34: I had concerns first time I read this config-mutating stuff, but once I actually used the CLI from the command line I can see that it is super useful. I might steal this pattern in future :).

client/checker_client/cli.py
```diff
@@ -0,0 +1,229 @@
...
+    "config_file",
+    default=CONFIG_FILE,
+    help="path to CLI config file",
+    show_default=True,
+)
+def cli(ctx, config_file):
+    """Checker command line utilities
+
+    By default this tool caches options for things like addresses in the
+    specified config file to make chaining calls easier.
```
utdemir, line 67: To me, the word "caches" was a bit misleading since it implies a performance optimisation. I would prefer something like "remembers".

Also, we might want to explain what exactly is going on a bit more; something like "the results of a command becomes the default arguments of the following commands" (with a nicer wording).

client/checker_client/cli.py
```diff
@@ -0,0 +1,229 @@
...
+from checker_client import checker as checker_lib
+import pytezos
+from pathlib import Path
+from dataclasses import dataclass
+from marshmallow import Schema, fields
+from marshmallow.decorators import post_load
+import portpicker
+import json
+
+CONFIG_FILE = Path.home().joinpath(Path(".checker"))
```
utdemir, line 12: Just to nitpick. We can look at "$XDG_CONFIG_HOME", or default to "$HOME/config" when it is not present. Turns out there is a package called `xdg` which does that, but I don't know if it is worth introducing a new dependency.

client/checker_client/cli.py
```diff
@@ -0,0 +1,229 @@
...
+@click.pass_obj
+def start(config: Config):
+    """Starts the sandbox"""
+    if checker_lib.is_sandbox_container_running(Config.sandbox_container):
+        click.echo(f"Container {Config.sandbox_container} is already running.")
+    else:
+        click.echo("Starting sandbox container...")
+        if not portpicker.is_port_free(config.sandbox_port):
+            click.echo(f"Selected host port {config.sandbox_port} for sandbox.")
+            config.sandbox_port = portpicker.pick_unused_port()
```
utdemir, line 94: I don't really understand this part. If the sandbox_port is not free, do we implicitly select a new one? This is a bit weird because if the user explicitly has made a choice we should respect that. Also, why are we printing the old port, then picking a new one?

Actually, since this command line only spins up one sandbox, the `portpicker` might be unnecessary and we can hardcode a port number.

utdemir (approved):
I love this PR. I just have a bunch of minor comments.

I am also thinking about whether having a `dev` subcommand and having sandbox related commands, and `deploy ctez` and `deploy mock_oracle` command under the `dev` subcommand would be better. But I am not sure and that's something we can always change later on.

purcell:
What a great review!

client/checker_client/cli.py
dorranh: Thanks! This was also my first time trying this. I also had some doubts, but it really is convenient. I think `gcloud` does something similar which is kind of where I got the idea from.

client/checker_client/cli.py
dorranh: Done. TIL about $XDG_CONFIG_HOME as well

client/checker_client/cli.py
dorranh: Good catch on the misplaced echo command, and that makes sense. Originally you actually couldn't explicitly specify the sandbox port, so I figured it would be easiest to let it pick one automatically. I agree that it makes sense to let people manually specify one if there is a conflict (since there is only one container at the moment anyways) and updated things accordingly.

PR #131, opened 2021-05-19
Refactor burrow slice list logic into separate module

dorranh:
Wanted to go ahead and get this up here to discuss before making any further changes, please let me know what you think!

This PR moves the logic for working with the double-linked lists formed by burrow slices our of `liquidationAuction.ml` into its own module. The goal of this refactor is to reduce the number of things going on in the various functions in `liquidationAuction.ml` in order to make them easier to test.

Note that I have kept this totally internal to make the refactoring a bit easier (and avoid changes to the existing tests), so the types stored in the auction state are the same as before. This adds a bit of extra work for the endpoints operating on the linked lists. I also noticed that the size of a couple of the endpoints increased by ~100-200 bytes and wanted to check if that was an issue given our current size and deployment approach.

The new `SliceList` module acts as a high-level interface to the AVL queue(s). I originally was going to make it only handle the list logic, but it turns out that we always need to update the burrow slice lists whenever we send or remove a slice from the AVL. The challenges in the seemingly complex functions like `take_with_splitting` actually boil down to retrieving the appropriate list and element from the AVL storage and are handled via a couple of constructors (e.g. `slice_list_from_queue_head`)

If we want to move forward with this solution I can do something similar for the completed auctions lists. I originally wanted to make the module in this PR polymorphic to avoid having a similar-looking module, but doing so in Ligo seems to be impossible as far as I can tell.


src/sliceList.ml
```diff
@@ -0,0 +1,195 @@
...
+    in
+    (* Touch up the old element in the backend *)
+    let former_youngest = bounds.slice_list_youngest_ptr in
+    let storage = avl_update_leaf storage former_youngest (fun (s: liquidation_slice) -> {s with younger = Some ptr}) in
+    (* Update up the bounds with the new youngest element *)
+    let bounds = {bounds with slice_list_youngest_ptr = ptr} in
+    storage, SliceList {meta with slice_list_bounds=Some bounds;}, SliceListElement (ptr, slice)
+
+(* Remove the element from the list, returning its contents *)
+let slice_list_remove (l:slice_list) (storage:mem) (e:slice_list_element) =
```
utdemir, line 132: I think I would appreciate annotating the return types at top-level functions. My first instinct to understand a function is usually looking at the type signature, but here it is missing half the information.

Interesting thing is that `ligo` used to require having full type signatures on functions, I guess that requirement is relaxed at one point.

src/sliceList.ml
```diff
@@ -0,0 +1,195 @@
...
+let slice_list_empty (burrow: burrow_id) = SliceList {slice_list_burrow=burrow; slice_list_bounds=(None:slice_list_bounds option);}
+
+let slice_list_is_empty (l: slice_list) =
+  let meta = match l with SliceList meta -> meta in
+  match meta.slice_list_bounds with
+  | Some _ -> false
+  | None -> true
+
+(* Constructs a burrow slice list for the given burrow id using the provided auction state *)
+let slice_list_from_auction_state (auctions: liquidation_auctions) (burrow_id: burrow_id) =
```
utdemir, line 44: If some of these functions are used at most once per entrypoint, we can add a `[@inline]` pragma to them which can sometimes reduce the codebase.

src/liquidationAuction.ml
```diff
@@ -578,57 +441,12 @@ let liquidation_auction_get_current_auction (auctions: liquidation_auctions) : c
...
-           assert (older.younger = Some leaf_ptr);
-           { older with younger = leaf.younger }
-        )
-  ) in
-
-  (* return *)
-  ( leaf.contents
+  let element, burrow_slices = slice_list_from_leaf_ptr auctions leaf_ptr in
+  let avl_storage, burrow_slices, root_ptr, contents = slice_list_remove burrow_slices auctions.avl_storage element in
+  let auctions = {auctions with avl_storage=avl_storage;} in
```
utdemir, line 446: I saw this pattern (we pass `avl_storage` and `slice_list` around, then put the `avl_storage` back into the auctions and call `slices_to_auction_state`) multiple times, maybe this means that `avl_storage` needs to be in `SliceList`?

src/avl.ml
```diff
@@ -441,6 +441,8 @@ let avl_push
       (mem, LeafPtr leaf_ptr)
   end
 
+type queue_end = Front | Back
```
utdemir, line 444: Is there a better place for this type given that it is not used anywhere in this module?

src/sliceList.ml
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+(* Constructs an element from a burrow leaf in the AVL *)
+let slice_list_from_leaf_ptr (auctions: liquidation_auctions) (ptr: leaf_ptr) =
+  let slice = avl_read_leaf auctions.avl_storage ptr in
+  let element = SliceListElement (ptr, slice) in
+  let list = slice_list_from_auction_state auctions slice.contents.burrow in
+  let _ = if slice_list_is_empty list then
+      failwith "invariant violation: corresponding list for slice was empty"
+    else ()
```
utdemir, line 63: Should this be `assert (not (slice_list_is_empty list));`? If we do that it'll be removed when transpiling to ligo.

utdemir:
I think this is better than the existing code. However to be honest I find both this module and the existing one error prone, there is just too many things that can be missed when using them.

I liked the idea of having a subset of auctions type focused to a specific burrow id (`SliceList`) and manipulating it; but it also suggests that maybe there is a different way of storing our state that makes this representation change unnecessary. I think this is what you meant here, right?:

>  so the types stored in the auction state are the same as before. This adds a bit of extra work for the endpoints operating on the linked lists.

---

>  I also noticed that the size of a couple of the endpoints increased by ~100-200 bytes and wanted to check if that was an issue given our current size and deployment approach.

I don't think it specifically is a big issue (although less would be better :) ). It's also possible that inlining some functions will make it better. But I am suprised that we didn't end up simplifying any redundant code paths.

---

In short, I think it is good to merge this one since it is better than the existing solution; but it still is probably the most brittle piece of checker. We got to find a way to simplify these even more.

src/liquidationAuction.ml
```diff
@@ -187,50 +187,10 @@ let liquidation_auction_send_to_auction
...
-            contents.burrow
-            new_burrow_slices
-            auctions.burrow_slices;
-      } in
-    assert_liquidation_auction_invariants new_state;
-    (new_state, ret)
+    let burrow_slices = slice_list_from_auction_state auctions contents.burrow in
+    let storage, burrow_slices, (SliceListElement (ret, _)) = slice_list_append burrow_slices auctions.avl_storage auctions.queued_slices Back contents in
+    let state = {auctions with avl_storage=storage} in
+    (slice_list_to_auction_state state burrow_slices, ret)
```
gkaracha, line 193: Might be a good idea to keep the `assert_liquidation_auction_invariants` here.

dorranh:
Thanks for the reviews! I'm going to move this PR to a draft state until we have more time to discuss (see #135) since I don't think it adds much in terms of test coverage, etc. in the immediate term.

src/avl.ml
dorranh: Ya at first I thought this might be useful in avl.ml, but since this makes sense. Moved it over to sliceList.ml for the time being.

src/sliceList.ml
dorranh: I actually wasn't sure on this point. I tried to mimic the behavior [in liquidationAuction.ml](https://github.com/tezos-checker/checker/blob/ff6259de4d47f7b4a2fe52aaca33e5c6884f4b2c/src/liquidationAuction.ml#L589) since if this case occurs the function will implode.

dorranh:
@utdemir @gkaracha, thanks again for the initial feedback the other day! I've made the request changes but am going to wait on @gkaracha's PR reducing the gas cost of sealing the contract before moving forward. That way I can test whether the changes in this PR exceed the gas limit.

dorranh:
It appears that this change blows up the gas cost of operations which interact with `liquidationAuction.ml` beyond the gas limit (sealContract now exceeds the gas limit when deploying). We shouldn't merge this until the gas costs have been reduced.

dorranh:
Alright, finally got the gas costs down! Here's a comparison with master

(this PR @ 8d34def0e6c90be9b3a4ff6107331f8fff078b63)
```
Gas costs:
{
    "touch": "541631",
    "remove_liquidity": "74707",
    "add_liquidity": "74076",
    "buy_kit": "70115",
    "sell_kit": "69237",
    "deactivate_burrow": "57355",
    "withdraw_tez": "55740",
    "set_burrow_delegate": "52038",
    "mint_kit": "51482",
    "deposit_tez": "50414",
    "create_burrow": "47993",
    "touch_burrow": "45015"
}
```

master @ a121393015ee82d5660167a426146767dd31749b

```
Gas costs:
{
    "touch": "530610",
    "remove_liquidity": "74707",
    "add_liquidity": "74076",
    "buy_kit": "70115",
    "sell_kit": "69240",
    "deactivate_burrow": "57355",
    "withdraw_tez": "55740",
    "set_burrow_delegate": "52038",
    "mint_kit": "51482",
    "deposit_tez": "50414",
    "create_burrow": "47993",
    "touch_burrow": "45015"
}
```

PR #133, opened 2021-05-21
Implement views suggested by FA2

utdemir:
Pretty much implementing the following: https://gitlab.com/tzip/tzip/-/blob/4b3c67/proposals/tzip-12/tzip-12.md#off-chain-views

* I had to fix the regex on generate-entrypoints since it didn't handle parameter type containing `*` correctly.
* I also noticed that previously I had implemented the operator logic in FA2 incorrectly, it turns out operators are per token_id. This commit also fixes that.

This PR omits the FA2 "token_metadata" section that we should implement separately.

@gkaracha, could you double check the "view_total_supply" function? I _think_ I am using correct values there, but I am wondering if there could be issues with rounding.

### DIFF
--- a/scripts/generate-entrypoints
+++ b/scripts/generate-entrypoints
@@ -9,7 +9,7 @@ ARGF.each_line do |line|
    case line
    when /^val +entrypoint_(\S+).*checker\s+\*\s+(.*?)\s+->/ then entrypoints << EntrypointInfo.new( $1, $2)
    when /^val +entrypoint_/ then raise "#{ARGF.filename}:#{ARGF.lineno}: Unrecognised entrypoint decl #{line}"
-   when /^val view_(\S+) *: *\(([^*]*)\*.*-> *(.*)/ then views << ViewInfo.new($1, $2.strip, $3.strip)
+   when /^val view_(\S+) *: *\((.*?) *\* *checker\) *-> *(.*)/ then views << ViewInfo.new($1, $2.strip, $3.strip)
    when /^val +view/ then raise "#{ARGF.filename}:#{ARGF.lineno}: Unrecognised view decl #{line}"
    end
 end
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -791,3 +791,33 @@ let view_is_burrow_overburrowed (burrow_id, state: burrow_id * checker) : bool =
 
 let view_is_burrow_liquidatable (burrow_id, state: burrow_id * checker) : bool =
   burrow_is_liquidatable state.parameters (find_burrow state.burrows burrow_id)
+
+(* ************************************************************************* *)
+(**                            FA2_VIEWS                                     *)
+(* ************************************************************************* *)
+
+let view_get_balance ((owner, token_id), state: (Ligo.address * fa2_token_id) * checker) : Ligo.nat =
+  fa2_get_balance (state.fa2_state, owner, token_id)
+
+let view_total_supply (token_id, state: fa2_token_id * checker) : Ligo.nat =
+  (* TODO: we should have assertions that the total amounts in parameters and cfmm is consistent with
+     what's in the ledger. Alternatively, we can use the ledger as a source of truth and remove
+     parameters.circulating_kit and cfmm.lqt.
+   *)
+  if token_id = kit_token_id then kit_to_mukit_nat state.parameters.circulating_kit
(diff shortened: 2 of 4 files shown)


scripts/generate-entrypoints
```diff
@@ -9,7 +9,7 @@ ARGF.each_line do |line|
    case line
    when /^val +entrypoint_(\S+).*checker\s+\*\s+(.*?)\s+->/ then entrypoints << EntrypointInfo.new( $1, $2)
    when /^val +entrypoint_/ then raise "#{ARGF.filename}:#{ARGF.lineno}: Unrecognised entrypoint decl #{line}"
-   when /^val view_(\S+) *: *\(([^*]*)\*.*-> *(.*)/ then views << ViewInfo.new($1, $2.strip, $3.strip)
+   when /^val view_(\S+) *: *\((.*?) *\* *checker\) *-> *(.*)/ then views << ViewInfo.new($1, $2.strip, $3.strip)
```
gkaracha, line 12: Woof, that's on me, I missed it while reviewing #132, sorry about that :confused:

src/fa2Interface.ml
```diff
@@ -232,21 +239,28 @@ let[@inline] fa2_run_update_operators
...
                  st.operators;
            }
     )
     st
     xs
 
+let[@inline] fa2_get_balance (st, owner, token_id: fa2_state * Ligo.address * fa2_token_id): Ligo.nat =
+  let ledger = st.ledger in
+  let key = (token_id, owner) in
+  let () = assert_valid_fa2_token token_id in
```
gkaracha, line 252: Note that it would have been safer to use "ensure" instead of "assert" (i.e., `ensure_valid_fa2_token`), but I saw now that our `generate-ligo.sh` script is not replacing _all_ `assert`-prefixed lines anymore so it should be OK either way :slightly_smiling_face:

gkaracha: Since you have this assertion as part of `fa2_get_balance`, perhaps you can now remove the same one in `fa2_run_balance_of` (line 262 below)?

gkaracha (approved):
LGTM! I left a few minor comments but otherwise I think it's good to go. Marking checker's kit on the ledger I think we should do, but it can wait for another PR.

src/checker.ml
```diff
@@ -791,3 +791,33 @@ let view_is_burrow_overburrowed (burrow_id, state: burrow_id * checker) : bool =
...
+(* ************************************************************************* *)
+
+let view_get_balance ((owner, token_id), state: (Ligo.address * fa2_token_id) * checker) : Ligo.nat =
+  fa2_get_balance (state.fa2_state, owner, token_id)
+
+let view_total_supply (token_id, state: fa2_token_id * checker) : Ligo.nat =
+  (* TODO: we should have assertions that the total amounts in parameters and cfmm is consistent with
+     what's in the ledger. Alternatively, we can use the ledger as a source of truth and remove
+     parameters.circulating_kit and cfmm.lqt.
+   *)
```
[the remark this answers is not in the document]
utdemir: Yeah, that'd be good. I guess then we need to modify ledger to store the total amount per token, and update all other functions to use those. Can you create an issue about it, since I am not 100% sure about the intricacies of `parameters.circulating_kit`?

scripts/generate-entrypoints
utdemir: Nah, those lines aren't meant to be read :).

src/fa2Interface.ml
utdemir: I agree with both comments. 

As about the first comment, I think we are overdue for a small rethink of `failwith`, `Ligo.failwith` and assertions. Probably soon, because TZIP-16 spec also has a field called `errors` that requires us to have structured error code to message mapping.

gkaracha: Right, good point. Our error codes though are still a little fluid so that would easily get out of sync. I certainly agree though that we gotta revisit the `failwith`/`assert` situation.

src/checker.ml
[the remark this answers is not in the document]
gkaracha: Yeah, I will open an issue and try to address this one, to make sure it agrees with the recent changes to the circulating_kit. :+1:

utdemir:
Thanks @gkaracha ! I applied your changes, and commented out the unimplemented token_metadata view which can be implemented on a separate PR, so I am merging this one.

PR #137, opened 2021-05-24
Introduce treefmt

utdemir:
I noticed there were some ocaml files where `ocp-indent` wasn't applied. I also noticed that @dorranh was passing the Python sources from an indenter, which is a good idea.

So, I wanted to ensure that our code is indented on the CI, while being able to indent Ocaml and Python at the same time (I also added `nixpkgs-fmt` because I could, but it's not super important). 

There's a new tool called `treefmt` that does that. So I added that to the CI, and set it up to use:

* `ocp-indent` for ocaml
* `black` for python
* `nixpkgs-fmt` for nix

I didn't enabled Black support yet because I didn't want to introduce conflicts to `dorranh/clientTimeouts` branch. I also didn't reformatted all our files since the diff becomes massive if I do that. Do run `treefmt` locally if you want to see which files change. I will apply it right before we merge this branch.

I also had to update the Nixpkgs pin to get a newer version of `treefmt`, but we were due for an update anyway.

And I know we haven't discussed this before, so please feel free to disagree with the PR. Alternative is that we don't introduce a new tool, add our formatters to `make format`, run that on CI and assert on `git diff` being clean.


utdemir:
Actually, let's scrap treefmt and just add a `format.sh` script. No need to add a new dependency. I'll update this PR.

utdemir:
Actually, let's just have a `scripts/format.sh` instead of `treefmt`, no need to have another dependency.

PR #138, opened 2021-05-24
Enforce formatting on CI and format more extensions

utdemir:
Replaces #137.

I noticed there were some ocaml files where ocp-indent wasn't applied. I also noticed that @dorranh was passing the Python sources from an indenter, which is a good idea.

So, I wanted to ensure that our code is indented on the CI, while being able to indent Ocaml and Python at the same time (I also added nixpkgs-fmt because I could, but it's not super important). So I just put them all inside a `format.sh` script and removed the pre-commit hook (not fun to maintain two separate formatting thingies).

That's using:

* `ocp-indent` for ocaml
* `black` for python
* `nixpkgs-fmt` for nix
* `dune format-dune-file` for dune files

I did not include the actual reformatting yet since the diff becomes massive if I do that. Do run treefmt locally if you want to see which files change. I will apply it right before we merge this branch.


purcell:
Nice! Should the pre-commit not now call format.sh? I'm wary of people wasting time pushing to CI only for it to fail on indenting.

utdemir:
It's a bit hard, because the script doesn't know about which changes are actually staged, and it can only apply the changes to the current directory, instead of checking if everything is indented. And I'm wary of changing files without an explicit request to do so.

I guess we can work on the script to make it optionally only check staged files without any change, I'll take a look if it's easy.

purcell:
Any idea what caused the build failure after those further changes?

utdemir:
> Any idea what caused the build failure after those further changes?

Turns out I've missed a few dependencies in shell.nix, thanks!

I fixed it, and the current error is expected; it's complaining that some of the files are not formatted correctly. I'm not fixing them right now because it makes the PR diff pretty big, I can do it right before we merge this.

purcell:
Nice, I think any one of us can feel free to reformat the repo on top of this PR and merge it, or even just merge it and then reformat on `main` soon afterwards.

gkaracha:
> Nice, I think any one of us can feel free to reformat the repo on top of this PR and merge it, or even just merge it and then reformat on `main` soon afterwards.

Yeah, I can take care of this in a bit. I'd use the PR format first, and then merge the PR, so that CI always stays green.

gkaracha:
Actually @purcell let's wait a little for #131 and my current attempt to reduce the gas cost of `touch` before formatting the whole codebase.

PR #139, opened 2021-05-25
Update client to better handle deployment to non-sandbox networks

dorranh:
Enables the client to deploy to remote nodes / real networks and fixes a variety of stability issues causing timeout errors, etc.

PR #140, opened 2021-05-26
FA2 Token Metadata

utdemir:
This PR implements the [token_metadata functionality of TZIP-12][token-metadata-docs].

The way it is implemented is via supplying a `token_metadata` off-chain view of a certain type, which essentially is a `(string, bytes) map` per token type (kit and liquidity in our case).

It would be easy to just hardcode the values in the code, but that'd force people to recompile checker in order to customize token metadata. So I ended up writing the view by hand at deployment time. This is not so tricky here because it pretty much is only static data and does not require looking at our view.

The user can supply a JSON file containing their overrides with `--token-metadata` parameter, but without it we have some minimal defaults which seems to work:

```
>>> client.contract("KT19YUWzmowLnu5rQdWBNqiKBfWcoYF9XxjQ").token_metadata[0]
Trying to fetch token 0 metadata from off-chain view
Trying to fetch contract metadata from `tezos-storage:m`
ContractTokenMetadata(name='kit', symbol='KIT', decimals=6, symbolPreference=None, thumbnailUri=None, artifactUri=None, displayUri=None, description=None, minter=None, creators=None, isBooleanAmount=None, formats=None, tags=None, contributors=None, publishers=None, date=None, blockLevel=None, type=None, genres=None, language=None, identifier=None, rights=None, rightsUri=None, externalUri=None, isTransferable=None, shouldPreferSymbol=None, attributes=None, assets=None)
```

I also tried using Temple Wallet and it seems to be able to recognize the metadata, so the format seems to be ok.

In future, we can rename the `token-metadata` parameter to `metadata` and make it also include [TZIP-16 contract metadata][contract-metadata-docs].

[token-metadata-docs]: https://gitlab.com/tzip/tzip/-/blob/4b3c67/proposals/tzip-12/tzip-12.md#token-metadata
[contract-metadata-docs]: https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#metadata-json-format

### DIFF
--- a/client/checker_client/checker.py
+++ b/client/checker_client/checker.py
@@ -7,15 +7,65 @@ import subprocess
 import tempfile
 import time
 from collections import namedtuple
+from typing import Optional, List
 
 import docker
 import pytezos
 import requests
 
+default_token_metadata = {
+  "kit": {
+     "name": "kit",
+     "symbol": "KIT",
+  },
+  "liquidity": {
+     "name": "liquidity",
+     "symbol": "LQDT",
+  }
+}
+
+def get_token_metadata_view_from_file(*, token_metadata_file: Optional[str], token_info):
+    # start with defaults
+    metadata = default_token_metadata
+
+    # insert the user overrides
+    if token_metadata_file:
+        with open(token_metadata_file) as f:
+           for tok, ms in json.load(f).items():
+               metadata[tok].update(ms)
+
+    # insert the required metadata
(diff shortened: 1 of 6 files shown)


gkaracha (approved):
Nice, LGTM! :ok_hand: @dorranh I'd merge it myself, but I am leaving it to you, in case there are conflicts with #139 to sort out.

PR #143, opened 2021-06-01
Updates to e2e tests

utdemir:
I was working on testing the CFMM functionality of checker.

To do that, I had to use the `ctez` contract to get some ctez. While doing that; I

* Found a bug on `ctez` contract's deployment script on our side (see the first commit).
* Reused the docker sandbox logic from `checker_client` library on e2e tests
* Slightly tweaked the parameters we pass to `inject` function to make them more reliable. Currently the tests work more often than they fail, but unfortunately there still are some intermitten failures. However it takes a really long time to try out a change, so I couldn't delve into them so much.

In the end, I ended up being able to mint some `ctez`, but didn't have time finish the test I want to write. I can do it tomorrow, but the PR as-is is mergeable state, so we might as well review & merge this one.

### DIFF
--- a/client/checker_client/checker.py
+++ b/client/checker_client/checker.py
@@ -391,7 +391,7 @@ def deploy_ctez(tz: PyTezosClient, ctez_dir, num_blocks_wait=100, ttl: Optional[
         fa12_ctez_storage = {
             "tokens": {"tz1Ke2h7sDdakHJQh8WX4Z372du1KChsksyU": 1},
             "allowances": {},
-            "admin": tz.key.public_key_hash(),
+            "admin": ctez.context.address,
             "total_supply": 1,
         }
         fa12_ctez = deploy_contract(
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -1,12 +1,15 @@
 import logging
 import os
+import json
 import time
 import unittest
+from pprint import pprint
 
 import docker
 import portpicker
 import pytezos
 import requests
+from pytezos.operation import MAX_OPERATIONS_TTL
 
 PROJECT_ROOT = os.path.join(os.path.dirname(__file__), "../")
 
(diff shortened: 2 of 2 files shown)


client/checker_client/checker.py
```diff
@@ -391,7 +391,7 @@ def deploy_ctez(tz: PyTezosClient, ctez_dir, num_blocks_wait=100, ttl: Optional[
         fa12_ctez_storage = {
             "tokens": {"tz1Ke2h7sDdakHJQh8WX4Z372du1KChsksyU": 1},
             "allowances": {},
-            "admin": tz.key.public_key_hash(),
+            "admin": ctez.context.address,
```
dorranh, line 394: Ah my bad on this one.

e2e/main.py
```diff
@@ -69,26 +51,50 @@ def test_e2e(self):
...
+                   num_blocks_wait=100,
+                   max_iterations=WAIT_BLOCK_ATTEMPTS,
+                   delay_sec=WAIT_BLOCK_DELAY
+                )
+            )
+            return ret
+
+        def call_checker_endpoint(name, param, amount=0):
+            ret = call_endpoint(checker, name, param, amount)
+            gas_costs[name] = ret["contents"][0]["gas_limit"]
```
dorranh, line 79: FYI - Since we are actually calling the endpoints we can use the OperationResult to get the `consumed_gas` (pretty sure that is the name of the field). I've noticed that it is sometimes quite a bit less than the estimated `gas_limit` when working with a fresh contract. 

The approach here is a more conservative estimate though, so I'm also happy leaving it as-is.

dorranh (approved):
Nice, LGTM!

e2e/main.py
utdemir: I would actually prefer using the real `consumed_gas` field, however, `ret` there does not contain a field like that.

I think it should be possible to find the operation group and extract the information from there, but I'm afraid that'd make this code a bit more complex. So I think we can merge this one, but I'd be happy to change it to `consumed_gas` if we find an easy way to get it.

client/checker_client/checker.py
utdemir: It happens, the tests catch it easily at least :)

purcell:
Lots of fiddly work here, great job

PR #144, opened 2021-06-02
Ensure the FA2-related entrypoints fail if Tezos.amount > 0

gkaracha:
I am not entirely sure about this one, to be honest. While adding tests I noticed that our FA2-related entrypoints do not fail if `Tezos.amount <> 0tez`, which is the common practice if no tez is expected, so I added the checks myself (plus unit tests). However, I couldn't find such a restriction in the [FA1.2/TZIP7](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-7/tzip-7.md) or the [FA2/TZIP12](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md) spec. For what it's worth, ctez which conforms to TZIP7 also performs [this check](https://github.com/tezos-checker/ctez/blob/8cc56059c4c6785e98e1b7d8d0267048fd17f5a7/fa12.mligo#L154-L156).

Size updates:

Because of the two strict entrypoints (`strict_entrypoint_transfer` and `strict_entrypoint_balance_of`) the size of the main contract goes up a little, and the lazy entrypoint `update_operators` increases slightly in size as well:
```
Main contract    : ~8875 bytes => ~8987 bytes
update_operators :  ~454 bytes =>  ~487 bytes
```
I've tested locally and the contract is still deployable though; gas consumption changes must be negligible as well.

### DIFF
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -849,12 +849,14 @@ let entrypoint_receive_price (state, price: checker * Ligo.nat) : (LigoOp.operat
 
 let strict_entrypoint_transfer (state, xs: checker * fa2_transfer list) : (LigoOp.operation list * checker) =
   assert_checker_invariants state;
+  let _ = ensure_no_tez_given () in
   let state = { state with fa2_state = fa2_run_transfer (state.fa2_state, xs) } in
   assert_checker_invariants state;
   (([]: LigoOp.operation list), state)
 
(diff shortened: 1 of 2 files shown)


utdemir (approved):
I think this is a good check to have, and I don't think they're harmful. These endpoints are pretty small, so I don't think we should worry too much about their gas cost.

gkaracha:
I agree that these checks are good to have, thanks for merging!

PR #145, opened 2021-06-03
Add big_map utility module to Python client library

dorranh:
This PR includes some of the more generally-useful content from my exploration into gas costs.

Adds a utility module to `checker_client` which provides some helpers for working with big_maps in pytezos. Specifically this library has functions for identifying keys in a contract's big_maps by scanning a chain for diffs and a supplementary method for rendering the contract's storage with the full content of the big_map. This is useful in cases where you want to call RPCs such as `run_code`, which require you to pass the full storage state.

### DIFF
--- a/client/checker_client/bigmap.py
+++ b/client/checker_client/bigmap.py
@@ -1,26 +1,16 @@
+"""Utilities for working with contracts which use big_maps
+"""
+
 from typing import Any, Dict, Optional, Set
-from pytezos import operation, pytezos
+
 from pytezos.client import PyTezosClient
-from pytezos.contract.call import ContractCall
 from pytezos.contract.interface import ContractInterface
 from pytezos.michelson.micheline import micheline_value_to_python_object
-from pytezos.michelson.types.adt import ADTMixin
-from pytezos.michelson.types.base import MichelsonType, Undefined
-from pytezos.michelson.types.pair import PairType
 from pytezos.michelson.types import ListType
-from pytezos.operation.group import OperationGroup
-from pytezos.rpc.helpers import ForgeOperationsQuery
-from pytezos.rpc.protocol import BlockQuery
+from pytezos.michelson.types.adt import ADTMixin
+from pytezos.michelson.types.base import Undefined
+from pytezos.michelson.types.big_map import BigMapType
 from pytezos.rpc.search import BlockSliceQuery
-from pytezos.michelson.types.big_map import BigMapType, big_map_diff_to_lazy_diff
-from checker_client import checker as checker_lib
-import json
-from pytezos.michelson.sections.storage import StorageSection
-from pytezos.operation.content import format_mutez
-from pprint import pprint
-
-# def skip_nones(**kwargs) -> dict:
-#     return {k: v for k, v in kwargs.items() if v is not None}
 
 
 def scan_bigmap_keys(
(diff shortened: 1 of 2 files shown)


client/checker_client/bigmap.py
```diff
@@ -0,0 +1,162 @@
...
+        # will not be rendered.
+        if not hasattr(michelson_data, "items") or isinstance(michelson_data, ListType):
+            # Only render items which are not "undefined". Pytezos uses same behavior internally.
+            if michelson_data is Undefined:
+                return Undefined
+            return michelson_data.to_python_object()
+        # Otherwise, this element is a container with child elements, recurse into them
+        if isinstance(michelson_data, ADTMixin):
+            # Because reasons, need to select the second element from this generator
+            child_iterator = map((lambda e: e[1]), michelson_data.iter_values())
```
gkaracha, line 152: Hahaha, nice :stuck_out_tongue: If there is any source for the "reasons", it might be a good idea to drop a link/explanation here.

client/checker_client/bigmap.py
```diff
@@ -0,0 +1,162 @@
...
+            return _render_bigmap(
+                michelson_data.field_name,
+                michelson_data.to_python_object(),
+                contract_bigmap_keys,
+                contract,
+            )
+        # This node is not a big map and has no children, just return it
+        # Currently this also catches lists, which means that bigmaps nested within a list
+        # will not be rendered.
+        if not hasattr(michelson_data, "items") or isinstance(michelson_data, ListType):
```
gkaracha, line 144: Hmmm, I'm not really familiar with the `ListType` so I don't know if there are any catches, but how come you don't recurse into lists too?

client/checker_client/bigmap.py
```diff
@@ -0,0 +1,162 @@
...
+                for op in op_group:
+                    op_result = op["metadata"]["operation_result"]
+                    # Skip ops which don't operate on the contract of interest
+                    if (
+                        op["kind"] == "origination"
+                        and contract_address not in op_result["originated_contracts"]
+                    ):
+                        continue
+                    if op["kind"] == "transaction" and op["destination"] != contract_address:
+                        continue
```
gkaracha, line 59: Do we want to check `op["source"]` here too, perhaps? Maybe not, since checker is originated :thinking:

gkaracha (commented):
Nice! I tried it locally and I can indeed see the contents of the bigmaps :ok_hand: I left a few minor comments/questions, but otherwise I'd say you could park it on master as-is. The approach is a little hacky to begin with and we don't need this functionality at the moment (I think) since the off-chain views have been fixed, so it's probably mostly useful for debugging purposes.

client/checker_client/bigmap.py
dorranh: Haha very true. In this case it is purely an API difference between the two iterators, so I'm not sure if it is worth documenting further.

client/checker_client/bigmap.py
dorranh: So in this case what will happen is the sub-elements of the list will get rendered, but a list of big_maps will not. I mainly limited this since I couldn't find a good example of how this would behave in pytezos and was also not sure whether it was even possible to store big_maps within a list in Michelson

client/checker_client/bigmap.py
dorranh: I think it is okay to leave out in this case. With the current implementation we also collect big_map diffs which occurred due to transactions besides those we sent, which I think might actually be desirable when using this outside of the sandbox environment.

dorranh:
Sounds good to me, thanks for the review @gkaracha!

client/checker_client/bigmap.py
gkaracha: True, big maps cannot appear in all contexts, but my impression is that this has been somewhat relaxed over the years so you never know! In any case, I assume that this is only for debugging purposes so it should be fine as is for now  :slightly_smiling_face:

PR #151, opened 2021-06-08
Long overdue fiddling in the tests folder

gkaracha:
I guess this PR is a little bikesheddy, but I've been wanting to perform these changes for a while now, and every now and then I come across these issues. Changes:

* Ensure that all calls to `assert_equal` get the optional `printer` argument, so that when there is a failure we can actually see what went wrong. Exception to this rule is the call to `assert_equal` in `test_all_permutations`, because it's very tight and it timeouts if we do.
* Instead of replicating the `assert_equal ~printer:(...) ~cmp:(...) expected real` pattern, define separate functions per type, like this:
  ```
  let assert_ratio_equal ~expected ~real = OUnit2.assert_equal ~printer:Ratio.show_ratio ~cmp:Ratio.eq_ratio_ratio expected real
  ```
  This has several advantages:
  - The code is shorter (this PR takes ~100 LOC away)
  - The order of `real` and `expected` cannot be gotten wrong (we had several instances where the arguments were flipped, giving misleading errors).
  - The use of the default `(=)` is avoided. If we don't pass the `cmp` parameter explicitly, `(=)` is implicitly used, which is wrong for several of our types, such as bigmaps, ratios, and every type that includes them. In fact, there are a few cases in the codebase were `(=)` is still used while it shouldn't, but I'll fix that in a separate PR.

### DIFF
--- a/tests/testAvl.ml
+++ b/tests/testAvl.ml
@@ -5,10 +5,9 @@ open Kit
 open OUnit2
 open Format
 open Ptr
+open TestCommon
 
 type auction_outcome_option = auction_outcome option [@@deriving show]
-type liquidation_slice_list = liquidation_slice list [@@deriving show]
-type slice_option = liquidation_slice option [@@deriving show]
 
 let nTez (i: int) : Ligo.tez = Ligo.tez_from_literal (string_of_int (1_000_000 * i) ^ "mutez")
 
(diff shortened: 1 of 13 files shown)


gkaracha:
@dorranh I hope this PR does interfere much with work in progress you might have :sweat_smile:

purcell (commented):
Looks tidier! 🥇

dorranh (approved):
Nice, LGTM! No worries with regards to any ongoing work on my end. If we end up being able to make #131 work it should not have any major conflicts since it mostly only touches a new test module which isn't affected by this PR.

gkaracha:
Cool, I'll go ahead and merge this one then, thanks!

PR #158, opened 2021-06-16
Update helpers for injecting operations

dorranh:
This PR adds (another) module to our Python client which implements updated versions of `wait_operations` and `inject` which better suit our deployment / DevOps use cases (i.e. they are more stable). It also switches the project back over to pytezos on PyPI from my fork.

I would prefer to avoid adding more code to our client library, but after all of the issues we've been running into with the default inject() method I think this is the best solution for the time being.

A nice benefit of these changes is that they allow for faster baking times with `time_between_blocks` now set to 1s. This enables e2e tests to run quite a bit faster. While the numbers will vary, in my case the before/after times were pretty dramatic:

Before: 687.729s
After: 181.772s

### DIFF
--- a/client/checker_client/operations.py
+++ b/client/checker_client/operations.py
@@ -2,6 +2,7 @@
 """
 
 import time
+from datetime import datetime
 from typing import Dict, List, Tuple
 
 from pytezos.client import PyTezosClient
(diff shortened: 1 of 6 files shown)


client/pyproject.toml
```diff
@@ -9,9 +9,7 @@ checker = 'checker_client.cli:cli'
 
 [tool.poetry.dependencies]
 python = "^3.8"
-# pytezos = "^3.1.0"
-pytezos = { git = "https://github.com/dorranh/pytezos.git" }
-# docker = "^5.0.0"
+pytezos = "^3.2.2"
```
gkaracha, line 12: Nice :sunglasses:

client/checker_client/operations.py
```diff
@@ -0,0 +1,164 @@
...
+    min_confirmations: int = 1,
+) -> Tuple[bool, List[dict]]:
+    """Awaits a list of operations to be included in a block.
+
+    Args:
+        tz: The pytezos client instance
+        op_hashes: A list of operation hashes to await
+        start_level: The starting level at which to start searching for operations. This should be a
+            level at or before the operations were injected.
+        min_confirmations: Minimum number of blocks required to consider operations as having been
```
gkaracha, line 46: Forgot to mention `max_blocks` here.

client/checker_client/operations.py
```diff
@@ -0,0 +1,164 @@
...
+) -> Dict:
+    """A replacement for pytezos's OperationGroup.inject
+
+    Submits the provided operation group and waits until it is confirmed by one block.
+
+    Args:
+        tz: pytezos client instance
+        op_group: The operation group to inject
+        max_wait_blocks: Maximum number of blocks to wait for a confirmation
+
```
gkaracha, line 138:
```suggestion

      min_confirmations: Minimum number of blocks required to consider operations as having been
            included on the chain.
```

gkaracha (commented):
Nice! A few tiny comments below; it LGTM overall :slightly_smiling_face:

utdemir (approved):
This worked well for me too. I tried it on my (yet uncommitted) stress tests and did not have any single problem even after repeated runs.

I think we should merge it, but still keep in touch with the upstream to get it as reliable too.

dorranh:
Excellent, I'm glad to hear that it worked. Thanks for the reviews guys! Just updated the docstrings and will merge once CI finishes.

PR #160, opened 2021-06-22
Enable E2E tests on the CI

utdemir:
After the changes on #158, our E2E tests seem to be not flaky. So this PR adds them to CI. We can always revert this if they start failing without reason.


purcell:
Yay, all this e2e work is great, CI appears to be passing so I'll merge now.

PR #162, opened 2021-06-22
Add bisect_ppx coverage annotations

dorranh:
Adds `bisect_ppx` coverage on / off annotations throughout the codebase to fix 1.) incorrect coverage reports which include call counts for type definitions due to interactions with `ppx_deriving` and 2.) unwanted coverage reporting for OCaml helpers which are only used in tests and not transpiled to Ligo. 

Follows the solution discussed here: https://github.com/aantron/bisect_ppx/issues/196.


src/ptr.ml
```diff
@@ -1,11 +1,15 @@
+[@@@coverage off]
 type ptr = Ligo.nat
 (* BEGIN_OCAML *) [@@deriving show] (* END_OCAML *)
+[@@@coverage on]
```
gkaracha, line 4: Well, all derivings are OCaml-only so I guess we don't need these `BEGIN_OCAML`/`END_OCAML` markers here. This looks like a good opportunity to get rid of them :wink: 
```suggestion
[@@@coverage off]
type ptr = Ligo.nat
[@@deriving show]
[@@@coverage on]
```

gkaracha (approved):
Flawless I believe! :ok_hand: 

Just to have the numbers here: with this PR coverage in the report increases from 85.24% to 88.35%.

However, while going through the coverage reports I realized that we don't process `checkerMain.ml` or `checkerEntrypoints.ml`, since neither of those files is managed by dune. I think we should eventually find a way to get a report for those too, but we can look into that in a separate PR.

Unimportant sidenote: Note that not all types between the coverage on/off markers derive show in `fa2Interface.ml` so we _could_ add more markers, to have coverage for the ones not deriving show at the moment. However, overall I think the way to go is to keep the markers where you placed them, and perhaps add those derivings later.

gkaracha:
By the way @dorranh while going through the coverage report I realized that `fraction_to_tez_ceil` and `fraction_to_nat_floor` are not used at all, but they bring down the numbers for `ratio.ml`. After we merge this PR I'll remove them—we might be able to pass 90% :slightly_smiling_face:

dorranh:
Thanks for the review! Good point about checkerMain.ml or checkerEntrypoints.ml - I created an issue for it (#163).

PR #164, opened 2021-06-23
Add a checker-level test for FA2 interface

utdemir:
I saw some lines of code on both fa2Interface.ml and checker.ml not covered by tests, so I ended up implementing a test suite.

### DIFF
--- a/tests/testChecker.ml
+++ b/tests/testChecker.ml
@@ -763,9 +763,7 @@ let suite =
              )
            ) in
 
-       (* initialize alice, bob and tom accounts *)
-       let tom_addr = Ligo.address_of_string "TOM_ADDR" in
-
+       (* initialize alice, bob and leena accounts *)
        Ligo.Tezos.new_transaction ~seconds_passed:10 ~blocks_passed:1 ~sender:initial_addr ~amount:(Ligo.tez_from_literal "0mutez");
        let _, checker = Checker.strict_entrypoint_transfer (checker, [
            { from_ = initial_addr;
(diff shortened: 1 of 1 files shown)


tests/testChecker.ml
```diff
@@ -736,6 +736,104 @@ let suite =
...
+
+       (* and tom can send some of that kit back to alice *)
+       Ligo.Tezos.new_transaction ~seconds_passed:10 ~blocks_passed:1 ~sender:tom_addr ~amount:(Ligo.tez_from_literal "0mutez");
+       let _, checker = Checker.strict_entrypoint_transfer (checker, [
+           { from_=bob_addr; txs=[{to_=alice_addr; token_id=kit_token_id; amount=Ligo.nat_from_literal "1n"}]}]) in
+
+       assert_nat_equal ~real:(balance checker alice_addr kit_token_id) ~expected:(Ligo.nat_from_literal "4n");
+       assert_nat_equal ~real:(balance checker bob_addr   kit_token_id) ~expected:(Ligo.nat_from_literal "1n");
+
+       (* but tom can not send rest of bob's kit back to alice when he's not an operator anymore *)
```
dorranh, line 821: We might be able to rephrase this to focus more on the second part, e.g. "but tom can not send even a single kit from bob's account when he's not the operator anymore"

dorranh (approved):
Nice! This scenario read very cleanly to me. Just one small comment on a comment, otherwise LGTM.

tests/testChecker.ml
```diff
@@ -736,6 +736,104 @@ let suite =
...
+           ( checker,
+             ( ctez_of_muctez (Ligo.nat_from_literal "5_000_000n")
+             , kit_of_mukit (Ligo.nat_from_literal "5_000_000n")
+             , lqt_of_denomination (Ligo.nat_from_literal "5n")
+             , Ligo.timestamp_from_seconds_literal 999
+             )
+           ) in
+
+       (* initialize alice, bob and tom accounts *)
+       let tom_addr = Ligo.address_of_string "TOM_ADDR" in
```
gkaracha, line 767: You might want to move this one into `test/testCommon.ml`, alongside `alice_addr`, `bob_addr`, and `leena_addr`, so that we don't have to redefine it in other places if we need more addresses.

gkaracha (approved):
More tests on FA2, nice!

tests/testChecker.ml
utdemir: Hey, I've missed the introduction of `leena`! I'll just use `leena_addr` instead of `tom`.

PR #166, opened 2021-06-23
Include checkerMain and checkerEntrypoints in the coverage report

### LINKED ISSUE #163: Code coverage for checkerMain and checkerEntrypoints
As noted by @gkaracha in #162, `checkerMain.ml` and `checkerEntrypoints.ml` don't get instrumented with `bisect_ppx` since the former isn't a part of the dune build and the latter is generated code. It would be great if we could either incorporate these into our existing tests to get coverage stats. This could perhaps be a good place to try out the Ligo test framework as well.

gkaracha:
Closes #163.

I got a little confused by this one, since `checkerMain.ml` and
`checkerEntrypoints.ml` are both already listed in `src/dune` after
all. Apparently `bisect_ppx` is a little lazy:
having `src/checkerEntrypoints.ml` and `src/checkerMain.ml` around
is not enough, but the testsuite must have _some_ dependency on
them for them to be included on the report. Simply `open`ing them
doesn't do it either. So, this PR:
- Adds basic tests for `CheckerMain.initial_wrapper`, to ensure
  that `checkerMain.ml` and `checkerEntrypoints.ml` are included
  in the test coverage report.
- Makes `make test-coverage` have an explicit dependency on
  `src/checkerEntrypoints.ml`, since the file is now needed for
  instrumentation.


gkaracha:
Locally I witness a test coverage reduction from 89.33% 83.25%.

dorranh (approved):
Nice, LGTM!

gkaracha:
Thanks for the review! :pray:

PR #170, opened 2021-06-25
Add additional unit tests

dorranh:
This PR adds a bunch of small, miscellaneous unit tests which were missing. Apologies that it got to be a bit large - I did my best to break the changes out into separate commits. As a part of these changes, `testCommon.ml` was renamed to `testLib.ml` to allow for a test module for `common.ml`. It also adds two calls in `generate-ligo.sh` which remove the other two possible coverage directives which we weren't previously using.

Note: This PR should also get `burrow.ml` up to 100% coverage once we can fix [the issue](https://github.com/aantron/bisect_ppx/issues/377) we encountered with `bisect_ppx`.


tests/testCommon.ml
```diff
@@ -1,101 +1,82 @@
-let alice_addr = Ligo.address_from_literal "alice_addr"
-let bob_addr = Ligo.address_from_literal "bob_addr"
-let leena_addr = Ligo.address_from_literal "leena_addr"
-let charles_key_hash = Ligo.key_hash_from_literal "charles_key_hash"
+open OUnit2
```
dorranh, line 1: Bah, it looks like the diff is rendering this a bit weirdly. This file is a new module, and `testLib.ml` is just the original `testCommon.ml` renamed.

utdemir (approved):
LGTM. I admit that some of the tests does not look as useful (eg. the ones testing the failure messages), but I do understand that they are necessary for getting the coverage up (and arguably better than adding a bunch more `[@coverage off]`'s.

dorranh:
Thanks, @utdemir! Totally agree. I wish there was a better way of handling these cases. I guess we could add more specific exception types though I'm not sure how much that makes sense since some of these errors are really "unexpected".

PR #174, opened 2021-06-29
Start adding tests for checkerEntrypoints.ml

gkaracha:
This PR includes only tests for the failing cases (i.e., trying to operate on the checker state while it's still unsealed). I wanted to add tests for the succeeding cases too, but this one got already big enough without them, so I'll open a separate PR for those cases.

Test coverage also increases (from 84.32% to 85.59%).

### DIFF
--- a/tests/dune
+++ b/tests/dune
@@ -19,6 +19,7 @@
   TestLiquidationAuction
   TestChecker
   TestCheckerMain
+  TestCheckerEntrypoints
   TestSliceList
   TestArbitrary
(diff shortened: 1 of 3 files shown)


purcell:
Nice! I pushed a further commit to reduce the duplication, since it appeared that the tests so far are all the same shape.

purcell:
Merging in the interest of generally landing stuff into `master` ASAP

gkaracha:
Great, thanks!

PR #176, opened 2021-06-29
Add some unit tests for cfmm views

gkaracha:
Test coverage also increases (from 84.32% to 85.14%).

### DIFF
--- a/tests/testChecker.ml
+++ b/tests/testChecker.ml
@@ -1311,6 +1311,71 @@ let suite =
        ()
     );
 
+    (* In order to (meaningfully) test the cfmm views, we need the cfmm to be
+     * populated. The following test reuses the same setup for testing all the
+     * cfmm views (otherwise we'd have to replicate this part per test). *)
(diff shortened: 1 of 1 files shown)


purcell:
I also pushed an extra commit here. PS. ounit2 also has a "bracket" combinator which might well be useful in some cases where there is common setup code: https://gildor478.github.io/ounit/ounit2/OUnit/index.html#bracket

gkaracha:
Thanks! Good tip for the bracket also, I didn't know of it but I'm pretty sure we have quite some setup, here and there. Might come in handy :+1:

purcell:
I forgot to mention that one general point here was something that I've seen in any number of languages with test frameworks: whenever you can refactor a test with multiple clauses into multiple tests, you potentially get more granular and useful failures. So although there's also a benefit to eliminating the possibility of consecutive clauses affecting each other via side effects, the granularity was what prompted me to add the refactoring commit.

PR #179, opened 2021-06-30
Increase test coverage for checkerMain.ml

gkaracha:
These unit tests primarily go through different execution
paths and ensure that calls to entrypoints fail/succeed,
depending on whether checker is sealed/unsealed.
However, the values themselves are not inspected.

At this high a level most of the actual code to be run is
wrapped in `{BEGIN,END}_LIGO`, so I think e2e tests
are much more appropriate for testing the behavior of
main.

Test coverage goes from 86.40% to 88.52%.

### DIFF
--- a/src/checkerMain.ml
+++ b/src/checkerMain.ml
@@ -52,46 +52,46 @@ let main (op, state: params * wrapper): LigoOp.operation list * wrapper =
   let ops, lazy_functions, metadata, deployment_state = match deployment_state with
     | Unsealed deployer ->
       begin match !Ligo.Tezos.sender = deployer with
-      | true ->
-        begin match op with
-        | DeployFunction p ->
-          let lfi, bs = p in
-          let lazy_functions =
-            match Ligo.Big_map.find_opt lfi lazy_functions with
-            | None -> Ligo.Big_map.add lfi bs lazy_functions
-            | Some prev -> Ligo.Big_map.add lfi (Ligo.Bytes.concat prev bs) lazy_functions in
-          (([]: LigoOp.operation list), lazy_functions, metadata, Unsealed deployer)
-        | DeployMetadata bs ->
-          let metadata =
-            match Ligo.Big_map.find_opt "m" metadata with
-            | None -> Ligo.Big_map.add "m" bs metadata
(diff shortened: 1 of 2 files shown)


dorranh (approved):
LGTM! I also agree on the point that detailed testing at this level should happen on either the end to end or ligo level. I see that you swapped one of the `if-else` statements for a `match-with`. Was the coverage not working properly in this case? I just ask because it's quite a long `if-else` statement with multiple function calls, so I would expect coverage to at least sort of work.

gkaracha:
Thanks for the review @dorranh! Indeed, I made two changes for coverage purposes:

- Moved the right-hand side of this clause to the same line as the pattern:
```diff
-         | CheckerEntrypoint _ ->
-          (Ligo.failwith error_ContractNotDeployed: LigoOp.operation list * lazy_function_map * (string, Ligo.bytes) Ligo.big_map * deployment_state)
+         | CheckerEntrypoint _ -> (Ligo.failwith error_ContractNotDeployed: LigoOp.operation list * lazy_function_map * (string, Ligo.bytes) Ligo.big_map * deployment_state)
```
  This at least makes the coverage report to mark the line as yellow (pattern: green, rhs: red due to https://github.com/aantron/bisect_ppx/issues/377), which I can read as green (if the pattern is matched, then the rhs is trivially accessed).

- Switched the `if-then-else` for a `match-with`, as you pointed out. I've noticed that the `else` is never marked red or green, so, to get the "yellow" behavior I mentioned in the previous point, I switched to pattern matching instead.

Overall, since I don't expect https://github.com/aantron/bisect_ppx/issues/377 to be changed any time soon I've been thinking that we can combine @aantron's [advice](https://github.com/aantron/bisect_ppx/issues/377#issuecomment-869589589) with using pattern matches instead of `if-then-else` to get accurate coverage. That is, if we convert
```ocaml
if <condition>
then e1
else (Ligo.failwith (...))
```
into
```ocaml
match <condition> with
| true -> e1
| false -> ((Ligo.failwith [@coverage off]) (...))        (* tricky *)
```
then the "tricky" line is colored green when covered by the tests and red when not covered by the tests, as we want.

PR #180, opened 2021-07-01
Increase test coverage for checkerEntrypoints.ml

gkaracha:
..via adding tests for all views `wrapper_view_*`. #174 dealt with
the failing cases, while this PR covers succeeding cases. To make
coverage more meaningful I used this opportunity to also tightly
check that the cfmm-related views calculate the maximum/minimum
values, as they are expected to do.

This PR brings test coverage up to 90.00%.


dorranh (approved):
LGTM!

gkaracha:
Awesome, thanks! :+1:

PR #181, opened 2021-07-05
Increase coverage for cfmm.ml

gkaracha:
Coverage increases

1. due to the addition of the following new tests:
   * cfmm_remove_liquidity - too much lqt withdrawn (equal)
   * cfmm_remove_liquidity - too much lqt withdrawn (more than)
   * cfmm_remove_liquidity - error_CfmmTooLate - barely (equal)
   * cfmm_remove_liquidity - error_CfmmTooLate - over (unequal)
   * view_buy_kit_min_kit_expected   - fail if no ctez is given
   * view_sell_kit_min_ctez_expected - fail if no kit is given
   * view_remove_liquidity_min_ctez_withdrawn - too much lqt withdrawn (equal)
   * view_remove_liquidity_min_ctez_withdrawn - too much lqt withdrawn (more than)
   * view_remove_liquidity_min_kit_withdrawn  - too much lqt withdrawn (equal)
   * view_remove_liquidity_min_kit_withdrawn  - too much lqt withdrawn (more than)
   * view_remove_liquidity_min_ctez_withdrawn - fail if no liquidity is given
   * view_remove_liquidity_min_kit_withdrawn  - fail if no liquidity is given
   * cfmm_add_liquidity - error_CfmmTooLate - barely (equal)
   * cfmm_add_liquidity - error_CfmmTooLate - over (unequal)
   * view_add_liquidity_max_kit_deposited - fail if no ctez is given
   * view_add_liquidity_min_lqt_minted    - fail if no ctez is given
2. due to the removal of the following inaccessible error paths (inaccessible because of arithmetic-related reasons)
   * `error_SellKitTooMuchTezBought`
   * `error_BuyKitTooMuchKitBought`
   * `error_AddLiquidityZeroKitDeposited`
   * `error_RemoveLiquidityTooMuchTezWithdrawn`
   * `error_RemoveLiquidityTooMuchKitWithdrawn`

Note that `cfmm.ml` has been and still is reported as having 100% test coverage by `bisect_ppx`, but before this PR it was at 92.31% real. ~~I think that with this PR we get to an actual 100%, but to check that I have to manually switch the `if-then-else`s to `match-with`s at the moment.~~ I (painstakingly) checked and we are now at real 100% coverage for `cfmm.ml`.

### DIFF
--- a/src/cfmm.ml
+++ b/src/cfmm.ml
@@ -29,11 +29,10 @@ let cfmm_kit_in_ctez_in_prev_block (cfmm: cfmm) =
  * cfmm contract was touched. *)
 let cfmm_sync_last_observed (cfmm: cfmm) : cfmm =
   assert (Ligo.geq_nat_nat !Ligo.Tezos.level cfmm.last_level);
-  match Ligo.leq_nat_nat !Ligo.Tezos.level cfmm.last_level with
-  | true ->
+  if Ligo.leq_nat_nat !Ligo.Tezos.level cfmm.last_level then
     (* do nothing if it's been touched already in this block *)
     cfmm
-  | false ->
+  else
     { cfmm with
       kit_in_ctez_in_prev_block =
         make_ratio
@@ -48,9 +47,9 @@ let cfmm_view_min_kit_expected_buy_kit
   : (kit (* min_kit_expected *) * cfmm) =
   let cfmm = cfmm_sync_last_observed cfmm in
   let cfmm = cfmm_assert_initialized cfmm in
-  match (ctez_amount = ctez_zero) with
-  | true -> (Ligo.failwith error_BuyKitNoTezGiven : (kit * cfmm))
-  | false ->
+  if (ctez_amount = ctez_zero) then
+    (Ligo.failwith error_BuyKitNoTezGiven : (kit * cfmm))
+  else
     (* db = da * (b / a) * (a / (a + da)) * (1 - fee) or
      * db = da * b / (a + da) * (1 - fee) *)
     let { num = num_uf; den = den_uf; } =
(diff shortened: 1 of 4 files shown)


gkaracha:
The removal of `error_SellKitTooMuchTezBought` and `error_BuyKitTooMuchKitBought` lead to a slight bytecode size reduction as well:
```diff
<                              buy_kit: ~1518 bytes, 1 chunks
>                              buy_kit: ~1472 bytes, 1 chunks

<                             sell_kit: ~1518 bytes, 1 chunks
>                             sell_kit: ~1472 bytes, 1 chunks
```

gkaracha:
Total bytecode size reduction:

```diff
<                              buy_kit: ~1518 bytes, 1 chunks
>                              buy_kit: ~1472 bytes, 1 chunks

<                             sell_kit: ~1518 bytes, 1 chunks
>                             sell_kit: ~1472 bytes, 1 chunks

<                        add_liquidity: ~2169 bytes, 1 chunks
>                        add_liquidity: ~2131 bytes, 1 chunks

<                     remove_liquidity: ~2311 bytes, 1 chunks
>                     remove_liquidity: ~2223 bytes, 1 chunks
```

src/cfmm.ml
```diff
@@ -120,15 +123,19 @@ let cfmm_view_min_ctez_expected_cfmm_sell_kit
...
-          kit = new_cfmm_kit;
-          ctez = ctez_sub cfmm.ctez bought_ctez;
-        }
-      )
+
+    (* Due to (a) the constant-factor calculation (which means that to deplete
+     * the one amount the other would in effect have to become infinite), (b)
+     * the fact that checker owns 1mu of each token, and (c) the fact that we
+     * always floor in our calculations, it should be impossible to trigger the
+     * following assertion. *)
```
utdemir, line 131: I feel like these repeated comments might get out of sync. I wonder if we should switch to something like GHC's `[Note]`'s, or just make one comment refer the other one. I guess it's not super important.

utdemir (approved):

src/cfmm.ml
gkaracha: Well, it's only two of them (because `buy_kit` and `sell_kit` are very much alike), so we should be OK. The other comments bellow (for `add_liquidity` and `remove_liquidity`) are different so there's not much sharing really. I do like the idea of GHC's `[Note]` though in general. We might be able to use it effectively in other places.

PR #182, opened 2021-07-06
ci: Publish code size and gas cost artifacts

utdemir:
This PR makes GitHub actions contain the code size and gas cost stats produced by the compilation script and the e2e test.

It is exposed as a `stats.zip` file that can be [found](https://github.com/tezos-checker/checker/actions/runs/1003859468) at the Actions page for a commit (Look at the second heading called "artifacts"). Unfortunately there is [no way](https://github.com/actions/upload-artifact#zipped-artifact-downloads) to have anything other than a zip file.

So, currently this is not extremely useful other than record-keeping purposes. But my plan is merging these first, and after a few weeks create a script to fetch last N artifacts to create a line graph. This might also make it possible to do fancier things like automatically sending a PR reporting the change. But we can think about them later on once we have more data.

Another thing we can do is to have another CI step (possibly nightly) running mutation tests, and export its results in a similar fashion. We can start thinking about it once our mutation-testing strategy is more refined.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -30,6 +30,17 @@ jobs:
         run: nix-shell --run 'make indent && if [ -n "$(git status --porcelain)" ]; then echo "Some files require formatting, run \"make indent\"."; exit 1; fi'
       - name: Build and test
         run: nix-build -A michelson --arg doCheck true
+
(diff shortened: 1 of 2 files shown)


gkaracha (approved):
Nice, LGTM! :ok_hand:

PR #183, opened 2021-07-06
Unit tests that catch mutations in cfmm.ml

gkaracha:
Tests added for the following mutations:
* src/cfmm.ml:245 (1_000_000 => 999_999)
* src/cfmm.ml:260 (kit_sub => kit_min)
* src/cfmm.ml:300 (kit_add => kit_max)


utdemir (approved):
LGTM, they look very clear.

gkaracha:
Great, thanks for the review!

PR #184, opened 2021-07-07
Parameterize mutate.py and build system changes

utdemir:
I wanted to have a nightly CI task to run our mutation script. In order to do that, I also had to teach dune to distinguish our fast and slow tests.

So, this PR contains a few changes:

* I changed our test suite to consist of a different dune task per module instead of a single task. I did that by removing the main entrypoint `test.ml`, and making each test suite their own dune target. So, now each one of our test suites can be invoked with a command like:

```
dune exec tests/testBurrow.exe 
```

* I added the aliases `run-fast-tests` and `run-avl-tests` to run a specific set of tests. Note that these do not yet cover all our test cases (eg. `testLiquidationAuction`), we should add them as we start mutation testing those modules. As before, `dune runtest` still runs all tests and it's what is used on the CI. 

You can use the aliases with:

```
dune build @run-fast-tests 
dune build @run-avl-tests

# or alternatively
make fast-test
```

This change is in the same spirit as https://github.com/tezos-checker/checker/pull/108. I choose to implement it as dune tasks more granularly, since I assume that the modules to test per mutation can overlap, and we might end up with arbitrary groupings.

* I modified our mutation script to take the list of modules and the command to test as parameter. Which allows us to invoke it like:

```
./scripts/mutate.py --test 'dune build @run-fast-tests' --num-mutations 2 parameters.ml burrow.ml
```

---

After this PR it should be easy to introduce a new cronjob (whatever the GitHub Actions equivalent is) to run something like the command right above. I haven't done this within this PR since it was getting too big.


src/dune
```diff
@@ -16,36 +16,6 @@
...
-  LiquidationAuctionPrimitiveTypes
-  Ptr
-  Ratio
-  BurrowTypes
-  Ligo
-  LigoOp
-  Fa12Interface
-  Fa2Interface
-  Error
-  Common)
```
utdemir, line 48: This is a sneaky change that is actually not necessary for this PR.

However, when I was reading the `dune` docs I saw that these are optional. `checker` indeed compiles without, so I went ahead and deleted them.

@dorranh I see that you changed bits and pieces here recently, can you think of a reason to specify the modules explicitly?

gkaracha (commented):
Nice, I think this is an improvement over what we currently have! :+1: I left a few comments about a couple of (seeming) omissions, but overall it LGTM. I still have to try it out locally though.

scripts/mutate.py
```diff
@@ -253,18 +251,43 @@ def do_mutation():
...
+
+parser.add_argument(
+    "--num-mutations", type=int, help="Number of mutations to test", default=25
+)
+
+
 if __name__ == "__main__":
-    n_mutations = 25
+    args = parser.parse_args()
+    random.seed(42)
```
dorranh, line 283: Random FYI - I don't think it matters in this case, but sometimes it can be useful to set this on the module level in case any random functions run on import or to avoid situations where you set the seed when calling main() but then forget to when importing a function from the module and calling it individually.

src/dune
dorranh: Nope, I think it should be fine without them. As far as I know, the only changes I've made here recently were to add the SliceList module.

dorranh (commented):
Nice, minus @gkaracha's comments this looks good to me and is much cleaner than the rough stab I took at this a couple of months ago :slightly_smiling_face: .

src/dune-project
```diff
@@ -1 +0,0 @@
-(lang dune 2.7)
```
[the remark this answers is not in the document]
utdemir: I just checked that it does not happen (at least on `make`). 

Actually, now the project does _not_ work with this file present:

```
$ dune test
File "tests/dune", line 24, characters 12-19:
24 |  (libraries checker core_kernel ounit2 qcheck)
                 ^^^^^^^
Error: Library "checker" not found.
Hint: try:
  dune external-lib-deps --missing @runtest
  ```

So maybe whatever was creating the file won't create it again? I'm not sure, we can always add it to `.gitignore` if it happens again.

File: scripts/format.sh [hunk 4]
```diff
@@ -11,4 +11,3 @@ $FD -e '.nix' -x nixpkgs-fmt
 $FD -e '.py' -x black
 
 new_dune="$(mktemp)"; dune format-dune-file src/dune > "$new_dune" && mv "$new_dune" src/dune
-new_dune="$(mktemp)"; dune format-dune-file tests/dune > "$new_dune" && mv "$new_dune" tests/dune
```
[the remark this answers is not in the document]
utdemir: No, I removed this on purpose.

I also prefer auto-formatting things, however in this case it makes every single line after https://github.com/tezos-checker/checker/pull/184/files#diff-f00bc4a85dc91983057a45088c3b4e6ac0d794731907dabd54d0293b71b22ed8R30 take 4 lines, removes the grouping between `run-avl-tests` and `run-fast-tests` and the entire thing does not fit into a single page anymore:

```
(rule
 (alias run-avl-tests)
 (action
  (run %{exe:testAvl.exe})))

(rule
 (alias run-avl-tests)
 (action
  (run %{exe:testAvlModel.exe})))

(rule
 (alias run-fast-tests)
 (action
  (run %{exe:testBurrow.exe})))

(rule
 (alias run-fast-tests)
 (action
  (run %{exe:testCommon.exe})))

(rule
 (alias run-fast-tests)
 (action
  (run %{exe:testCtez.exe})))

...
```

This all makes it much harder to follow what's going on. So I feel like it's a good trade-off here. Do you still think we should have it formatted?

scripts/mutate.py
utdemir: Yeah, my idea here was to make it a command line argument actually. Likely when we put this into CI we might want to run it on multiple threads/times with different seed values. What would you do in that case? I guess we can always make it something like an environment variable, but that does not look as clean.

dorranh: You know, I hadn't really considered this in the context of randomized testing before :thinking:. I have mostly used seeds to work with random quantities for which you want to reproduce a single result (models, unit tests for some statistics quantity). I guess a question I have would be how much value running the script from multiple, fixed seeds has over increasing the sample size of a single run. Would this just be to improve performance (potentially more unique mutations for a fixed `n_mutations` or roughly the same number of mutations for less execution time) or is the goal to increase the sample size?

In the case of the latter, I would just crank up the `n_mutations` in a single process and leave the seed at the module level. I would also do the same if time isn't an issue and you just want to maximize coverage since running with different seeds doesn't necessarily mean that sample set you get with each different seed will be different. 

However, if performance in terms of CI run-time is the issue then I think you're approach of providing the seeds via the command line makes good sense since the mutations on each thread will probably at least have some differences (especially if the sample size is low relative to the total number of possible mutations).

dorranh: PS: Another option for running on multiple threads to get more samples for less wall time would be to pre-calculate the mutations with a single seed for a large `n_mutations` then multi-thread the actual application of the mutations (which would require creating copies of the repo for each thread).

scripts/format.sh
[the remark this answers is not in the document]
gkaracha: Ah, I see. I agree then, we should be OK without formatting this one. :+1:

src/dune-project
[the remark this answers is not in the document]
gkaracha: > Actually, now the project does _not_ work with this file present:

Hmmm, that might be what I have a vague recollection of. Perhaps it's different on mac or something. For now we can merge this as is, and if anything breaks we can always revert it :smile:

scripts/mutate.py
gkaracha: Interesting topic, this one. Personally I'd be fine either way, but I think that I prefer getting all the parameters from the command line. Then, the mutation testing tool looks more like a pure function, from a CI point of view.

> I guess a question I have would be how much value running the script from multiple, fixed seeds has over increasing the sample size of a single run. Would this just be to improve performance (potentially more unique mutations for a fixed `n_mutations` or more mutations for less wall time) or is the goal to increase the sample size?

This I am not sure about, but I have some feedback on the usability. Whcn checking `cfmm.ml`, to get past the "noise" (i.e., meaningless mutations which started accumulating at the top of the report) I ended up having to raise `n_mutations` up to 2000, which took a very long time to run. Locally I could also avoid this by keeping `n_mutations` low and changing the seed manually a few times, and to replicate this behavior on CI I'd expect the cleanest way to be to pass the seed via the command line.

utdemir: This branch was bit-rotting, so I merged it. I'm not sure about how to do this to be honest, I also think being able to provide a seed value would be useful, on the other hand @dorranh 's concern makes sense too. 

I guess we have to see whether we get the same results when we provide the same seed.

PR #185, opened 2021-07-07
Add additional unit tests for burrow logic

dorranh:
This PR adds a variety of unit tests to `testBurrow.ml` which address missing test cases found during mutation testing:

```
SRC                  TYPE                 MUTATION_FROM   -> MUTATION_TO     LINE 
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> -1              L365  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 1_000_001       L365  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 999_999         L372  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> -1              L372  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 1               L372  
	
src/burrow.ml        INTEGER_LITERAL      0mutez          -> 1mutez          L210  

src/burrow.ml        INTEGER_LITERAL      0               -> -1              L538  
src/burrow.ml        SWAP_FUNCTION        gt_int_int      -> leq_int_int     L538  
src/burrow.ml        INTEGER_LITERAL      0               -> 1               L538  
src/burrow.ml        SWAP_FUNCTION        gt_int_int      -> eq_int_int      L538  
	
src/burrow.ml        SWAP_FUNCTION        add_int_int     -> div_int_int     L353
src/burrow.ml        SWAP_FUNCTION        add_int_int     -> mul_int_int     L353
src/burrow.ml        SWAP_FUNCTION        add_int_int     -> sub_int_int     L353  
	
src/burrow.ml        SWAP_FUNCTION        sub_int_int     -> div_int_int     L331  
src/burrow.ml        SWAP_FUNCTION        sub_int_int     -> mul_int_int     L337  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 999_999         L339  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> -1              L339  

src/burrow.ml        INTEGER_LITERAL      0mutez          -> 1mutez          L104  

src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 1               L438  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 0               L438  

src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 1               L409  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 999_999         L409  
src/burrow.ml        INTEGER_LITERAL      1_000_000       -> 0               L409  

```

### DIFF
--- a/tests/testBurrow.ml
+++ b/tests/testBurrow.ml
@@ -752,9 +752,6 @@ let suite =
             (Ligo.mul_int_int lp_den f_den))
     );
 
-    (* TODO: Property "compute_tez_to_auction - <= 0 for underburrowed burrows" *)
-    (* TODO: Property "compute_tez_to_auction - > 0 for overburrowed burrows" *)
-
     ("compute_tez_to_auction - expected value for an overburrowed burrow" >::
      fun _ ->
        let burrow = Burrow.make_burrow_for_test
--- a/tests/testBurrow.ml
+++ b/tests/testBurrow.ml
@@ -627,7 +627,7 @@ let suite =
          ~real:(Burrow.burrow_address  burrow)
     );
 
-    ("compute_min_kit_for_unwarranted - burrow with zero collateral and zero kit" >::
+    ("compute_min_kit_for_unwarranted - burrow with zero collateral and zero outstanding kit" >::
      fun _ ->
        let burrow0 = make_test_burrow
            ~outstanding_kit:kit_zero
@@ -641,7 +641,7 @@ let suite =
          ~real:min_kit_for_unwarranted
     );
 
-    ("compute_min_kit_for_unwarranted - burrow with zero collateral and positive kit" >::
+    ("compute_min_kit_for_unwarranted - burrow with zero collateral and positive outstanding kit" >::
      fun _ ->
        let burrow0 = make_test_burrow
            ~outstanding_kit:(kit_of_mukit (Ligo.nat_from_literal "1n"))
@@ -655,7 +655,7 @@ let suite =
          ~real:min_kit_for_unwarranted
     );
 
-    ("compute_min_kit_for_unwarranted - burrow with positive collateral and zero kit" >::
+    ("compute_min_kit_for_unwarranted - burrow with positive collateral and zero outstanding kit" >::
      fun _ ->
        let burrow0 = make_test_burrow
            ~outstanding_kit:kit_zero
@@ -669,7 +669,7 @@ let suite =
          ~real:min_kit_for_unwarranted
     );
 
-    ("compute_min_kit_for_unwarranted - burrow with positive collateral and positive kit" >::
+    ("compute_min_kit_for_unwarranted - burrow with positive collateral and positive outstanding kit" >::
      fun _ ->
        let burrow0 = make_test_burrow
            ~outstanding_kit:(kit_of_mukit (Ligo.nat_from_literal "1n"))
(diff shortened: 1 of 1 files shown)


tests/testBurrow.ml
```diff
@@ -644,6 +752,46 @@ let suite =
             (Ligo.mul_int_int lp_den f_den))
     );
 
+    (* TODO: Property "compute_tez_to_auction - <= 0 for underburrowed burrows" *)
+    (* TODO: Property "compute_tez_to_auction - > 0 for overburrowed burrows" *)
```
gkaracha, line 756: I'd hedge or even remove these two comments until we are sure that these properties hold (or not).

tests/testBurrow.ml
```diff
@@ -644,6 +752,46 @@ let suite =
...
+           ~excess_kit:kit_zero
+           ~active:true
+           ~address:alice_addr
+           ~delegate:None
+           ~collateral:(Ligo.tez_from_literal "3mutez")
+           ~adjustment_index:fixedpoint_one
+           ~collateral_at_auction:(Ligo.tez_from_literal "3mutez")
+           ~last_touched:(Ligo.timestamp_from_seconds_literal 0) in
+       (* Note: cranking up the index to make test more sensitive to small changes potentially obscured by rounding *)
+       let parameters = Parameters.({initial_parameters with index=(Ligo.tez_from_literal "100_000_000_000mutez");}) in
```
gkaracha, line 771: Hmmm, setting the index directly like this (instead of by touching the parameters for example) makes the difference between the index and the protected index somewhat "impossible" under normal circumstances:
```ocaml
index           = Ligo.tez_from_literal "100_000_000_000mutez"
protected_index = Ligo.tez_from_literal       "1_000_000mutez"
```
I guess it's OK since we're writing this test only to address the mutation, but in general setting fields in an ad-hoc way like this can lead to cases that are otherwise impossible to trigger.

Perhaps add a comment here about the distance between the index and the protected index for our future sanity :sweat_smile:

tests/testBurrow.ml
```diff
@@ -741,6 +889,102 @@ let suite =
...
+            ~delegate:None
+            ~collateral:(Ligo.tez_from_literal "2_470_000mutez")
+            ~adjustment_index:fixedpoint_one
+            ~collateral_at_auction:(Ligo.tez_from_literal "3_000_000mutez")
+            ~last_touched:(Ligo.timestamp_from_seconds_literal 0) in
+
+        assert_bool
+          "burrow_is_liquidatable returned true, but the burrow is expected to be non-liquidatable"
+          (not (Burrow.burrow_is_liquidatable Parameters.initial_parameters burrow))
+    );
```
gkaracha, line 949: Ha, I'm very surprised we need those two to be honest! Given how many times `burrow_is_liquidatable` is called in `tests/testLiquidation.ml` I'd have thought that it's pretty well-tested. Having more tests is always good I guess, if they cover previously uncovered territory :slightly_smiling_face:

gkaracha (approved):
LGTM! Left only a few minor comments (more like points for discussion, really) below.

tests/testBurrow.ml
gkaracha: Thinking more about this, I guess we can get the index to be very far from the protected index, if the oracle is compromised. Perhaps setting it explicitly is not that bad in this sense after all.

tests/testBurrow.ml
dorranh: Good eye, I had actually thought I had already removed these.

tests/testBurrow.ml
dorranh: I was as well! If I remember correctly, increasing the tez literal in this function (which corresponds to the tez scaling factor) _did_ get caught by our existing tests, but they didn't detect when it was set to a smaller value.

PR #187, opened 2021-07-08
Tests for parameters.ml, based on mutation testing results

gkaracha:
Tofix
---
Mutations relating to initial parameters.
- [x] `src/parameters.ml    INTEGER_LITERAL      1_000_000mutez  -> 999_999mutez    L29`

Mutations relating to adding/removing outstanding/circulating kit.
- [x] `src/parameters.ml    SWAP_FUNCTION        kit_sub         -> kit_add         L480`
- [x] `src/parameters.ml    SWAP_FUNCTION        kit_sub         -> kit_min         L480`
- [x] `src/parameters.ml    SWAP_FUNCTION        kit_sub         -> kit_add         L495`
- [x] `src/parameters.ml    SWAP_FUNCTION        kit_sub         -> kit_max         L495`

Mutations relating to the calculation of `q_{i+1}`. To address these we probably need tests changing `q` non-trivially.
- [x] `src/parameters.ml    INTEGER_LITERAL      6               -> 1               L295`
- [x] `src/parameters.ml    INTEGER_LITERAL      6               -> 0               L295`
- [x] `src/parameters.ml    INTEGER_LITERAL      6               -> -1              L295`
- [x] `src/parameters.ml    INTEGER_LITERAL      6               -> 5               L295`
- [x] `src/parameters.ml    INTEGER_LITERAL      6               -> 7               L295`
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> 1               L301`
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> -1              L301`
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> 0               L301`
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> 3               L301`

Mutations relating to the calculation of `drift_{i+1}`. To address these we probably need tests changing `drift` non-trivially.
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> 0               L265`
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> 1               L265`
- [x] `src/parameters.ml    INTEGER_LITERAL      2               -> -1              L265`

Mightfix
---
These mutations relate to `compute_drift_derivative`. However, they relate to conditions imposed on intermediate values, so it is unclear whether they can arise. More investigation needed.

**Edit: Wontfix. See comment below.**

- [x] `src/parameters.ml    SWAP_FUNCTION        lt_int_int      -> leq_int_int     L177`
- [x] `src/parameters.ml    SWAP_FUNCTION        lt_int_int      -> leq_int_int     L178`
- [x] `src/parameters.ml    SWAP_FUNCTION        leq_int_int     -> lt_int_int      L181`
- [x] `src/parameters.ml    SWAP_FUNCTION        gt_int_int      -> geq_int_int     L183`
- [x] `src/parameters.ml    SWAP_FUNCTION        geq_int_int     -> gt_int_int      L184`
- [x] `src/parameters.ml    SWAP_FUNCTION        leq_int_int     -> lt_int_int      L186`

Wontfix
---
**Own assertions**
- [x] `src/parameters.ml    SWAP_FUNCTION        lt_int_int      -> leq_int_int     L105`
- [x] `src/parameters.ml    SWAP_FUNCTION        gt_tez_tez      -> geq_tez_tez     L226`
- [x] `src/parameters.ml    INTEGER_LITERAL      0               -> -1              L428`
- [x] `src/parameters.ml    INTEGER_LITERAL      0mutez          -> 1mutez          L226`

**Impossible to trigger**
The following cannot be triggered, since they refer to the result of `kit_to_mukit_int` which takes a `nat` as an argument, so it always returns a non-negative number. Within this domain, `eq_int_int` and `leq_int_int` are interchangeable.
- [x] `src/parameters.ml    SWAP_FUNCTION        eq_int_int      -> leq_int_int     L89`
- [x] `src/parameters.ml    SWAP_FUNCTION        eq_int_int      -> leq_int_int     L90`
- [x] `src/parameters.ml    SWAP_FUNCTION        eq_int_int      -> leq_int_int     L92`

**Not worth addressing**
The following refer to inequalities; there is always a different number that the input is not equal to, so these are impossible to really satisfy. I have at least checked that we have many tests for this code path.
- [x] `src/parameters.ml    INTEGER_LITERAL      0               -> -1              L93` 
- [x] `src/parameters.ml    INTEGER_LITERAL      0               -> 1               L93`


scripts/mutate.py
```diff
@@ -131,9 +131,13 @@ def mutate_integer_literal(value_src: str) -> str:
         if value == 0:
             mutated_value = 1
         else:
-            mutated_value = random.choice([0, 1, value + 1, value - 1])
+            options_list = [0, 1, value + 1, value - 1]
```
dorranh, line 134: Doh, I was in the process of submitting a PR for this but you beat me to it :+1:. I can still submit a brief one including the logic for skipping assertions though.

gkaracha: Ah yeah, no worries! I applied a quick patch so that I can get rid of the numerous false alarms while working on this PR, but I'll revert it before merging. Feel free to open a separate PR :smile:

gkaracha:
By looking at each individual mutation in the Mightfix category, I can see now that these are also impossible to trigger. Let's take for example the mutation on line `parameters.ml:177`:
```ocaml
   Ligo.lt_int_int mul_sub_den_tlb_num_tlb_sf mul_target_tlb
-> Ligo.leq_int_int mul_sub_den_tlb_num_tlb_sf mul_target_tlb
```
If we inline all the numbers the condition ends up looking like this:
```ocaml
((1000 - 5) * 18446744073709551616) < (target * 1000)
```
which is equivalent to
```ocaml
18354510353341003857920 < target * 1000
```
There is no (natural) value for `target` that could make the two sides equal. The two sides can only be equal for
```ocaml
target = 18354510353341003857.92
```
but that's not a natural number.

That being said, I see no way to add tests for catching these mutations so I'm marking them as wontfix too.

gkaracha:
@dorranh this is good to go from my end, feel free to have a look now :slightly_smiling_face:

dorranh (approved):
Nice, LGTM! Just for general knowledge: does this cover most of the mutations in `parameters.ml`, or are there still some which need tests?

gkaracha:
@dorranh It should address all the mutations under "Tofix" above. The list (including the wontfix items) in the PR description includes all the mutations that were reported from `mutate.py` with 
```python
MODULES = ["parameters.ml"]
```
and
```python
n_mutations = 1000
```
It could be the case that increasing `n_mutations` might give more candidates, but I didn't try (it takes a while to run so many mutations).

PR #188, opened 2021-07-08
Reduce false positives generated by mutate.py

### LINKED ISSUE #186: Generated integer mutations in mutate.py are not always actual mutations
As noted by @gkaracha, sometimes mutate.py will generate "mutations" which are actually just an identity transformation (e.g. mutating `Ligo.int_from_literal "0"` to `Ligo.int_from_literal "0"`. This occurs because the current implementation selects possible mutations from a list, not accounting for the fact that input integer might be equal to one of the mutations in the list. We should update mutate.py to filter out these cases.

https://github.com/tezos-checker/checker/blob/d05caf304c816dcb6c4a3cc3662adca4383b7823/scripts/mutate.py#L136

dorranh:
This PR reduces the false positive rate of `mutate.py` by filtering out lines containing assertions and removing any mutation candidates for integer literal values which are equal to the original value. Fixes #186.

### DIFF
--- a/scripts/mutate.py
+++ b/scripts/mutate.py
@@ -131,10 +131,11 @@ def mutate_integer_literal(value_src: str) -> str:
         if value == 0:
             mutated_value = 1
         else:
-            mutated_value = random.choice([0, 1, value + 1, value - 1])
+            choices = [c for c in [0, 1, value + 1, value - 1] if c != value]
+            mutated_value = random.choice(choices)
     else:
-        mutated_value = random.choice([-1, 0, 1, value + 1, value - 1])
-
+        choices = [c for c in [-1, 0, 1, value + 1, value - 1] if c != value]
+        mutated_value = random.choice(choices)
     return f"{mutated_value:_}{suffix}"
 
 
--- a/scripts/mutate.py
+++ b/scripts/mutate.py
@@ -36,7 +36,7 @@ class MutationType(Enum):
     INTEGER_LITERAL = auto()
 
 
-MODULES = ["burrow.ml"]  #
+MODULES = ["burrow.ml", "checker.ml", "parameters.ml", "cfmm.ml"]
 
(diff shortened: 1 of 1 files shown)


scripts/mutate.py
```diff
@@ -162,6 +163,11 @@ def mutate(
         # seem to happen in the checker codebase though...
         if stripped.startswith("(*") or stripped.startswith("*"):
             continue
+        # Super basic logic for ignoring assertions. This might be overly
+        # aggressive (e.g. in cases with an inline comment talking about assertions), but
+        # this doesn't seem to be a common case.
+        if "assert" in stripped:
+            continue
```
gkaracha, line 170: I think this is probably OK as-is for now, but if you really want to match the way assertions are treated at the moment by the rest of the pipeline you can see if `stripped` matches the regexps we have in [`generate-ligo.sh`](https://github.com/tezos-checker/checker/blob/ad697ff76c97489d218265b91810df88c3ab74d7/scripts/generate-ligo.sh#L77-L82):
```
    ...
    # delete assertions
    sed 's/^ *assert .*//g' |
    sed 's/^ *assert_burrow_invariants .*//g' |
    sed 's/^ *assert_checker_invariants .*//g' |
    sed 's/^ *assert_liquidation_auction_invariants .*//g' |
    sed 's/^ *assert_burrow_slices_invariants .*//g' |
    ...
```

gkaracha (approved):
Nice! This is sorely needed :pray: I left a comment about assertions, but I think you can merge the changes either way.

scripts/mutate.py
dorranh: Good point, that makes sense. Since the cruder approach is actually a bit faster than using a regex I might leave it as-is for now since I don't think it is excluding too many (if any) legitimate lines which should be tested.

PR #191, opened 2021-07-12
Represent Parameters.{index,protected_index} as a Ligo.nat

gkaracha:
Until now we've been converting the `nat` we get from the
oracle into `tez` to match intuition, but in Michelson `tez`
is represented using 64 bit signed integers ([source](https://gitlab.com/tezos/tezos/-/blob/master/docs/009/michelson.rst)), which
can lead to unnecessary overflows.

### DIFF
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -744,7 +744,7 @@ let rec touch_oldest
  * exceptions (1. setting the delegate, and 2. call/callback to the oract), all
  * of the operations are outwards calls, to other contracts (no callbacks). It
  * should be safe to leave the order of the transaction reversed. *)
-let[@inline] touch_with_index (state: checker) (index:Ligo.tez) : (LigoOp.operation list * checker) =
+let[@inline] touch_with_index (state: checker) (index: Ligo.nat) : (LigoOp.operation list * checker) =
   assert_checker_invariants state;
   let
     { burrows = state_burrows;
@@ -834,7 +834,7 @@ let[@inline] touch_with_index (state: checker) (index:Ligo.tez) : (LigoOp.operat
 let entrypoint_touch (state, _: checker * unit) : (LigoOp.operation list * checker) =
   let index = match state.last_price with
     | None -> state.parameters.index (* use the old one *)
-    | Some i -> Ligo.mul_nat_tez i (Ligo.tez_from_literal "1mutez") in (* FIXME: Is the nat supposed to represent tez? *)
+    | Some i -> i in (* FIXME: Is the nat supposed to represent tez? *)
   touch_with_index state index
 
 (* ************************************************************************* *)
--- a/src/parameters.ml
+++ b/src/parameters.ml
@@ -8,8 +8,8 @@ open Constants
 
 type parameters =
   { q : fixedpoint; (* 1/kit, really *)
-    index: Ligo.tez;
-    protected_index: Ligo.tez;
+    index: Ligo.nat;
+    protected_index: Ligo.nat;
     target: fixedpoint;
     drift_derivative: fixedpoint;
     drift: fixedpoint;
@@ -26,8 +26,8 @@ type parameters =
 (** Initial state of the parameters. *)
 let initial_parameters : parameters =
   { q = fixedpoint_one;
-    index = Ligo.tez_from_literal "1_000_000mutez";
-    protected_index = Ligo.tez_from_literal "1_000_000mutez";
+    index = Ligo.nat_from_literal "1_000_000n";
+    protected_index = Ligo.nat_from_literal "1_000_000n";
     target = fixedpoint_one;
     drift = fixedpoint_zero;
     drift_derivative = fixedpoint_zero;
@@ -39,21 +39,21 @@ let initial_parameters : parameters =
   }
 
 (** Compute the current minting index (in tez). To get tez/kit must multiply with q. *)
-let[@inline] tz_minting (p: parameters) : Ligo.tez = max_tez p.index p.protected_index
+let[@inline] tz_minting (p: parameters) : Ligo.nat = max_nat p.index p.protected_index
 
 (** Compute the current liquidation index (in tez). To get tez/kit must multiply with q. *)
-let[@inline] tz_liquidation (p: parameters) : Ligo.tez = min_tez p.index p.protected_index
+let[@inline] tz_liquidation (p: parameters) : Ligo.nat = min_nat p.index p.protected_index
 
 (** Current minting price (in tez/kit). *)
 let minting_price (p: parameters) : ratio =
   make_ratio
-    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (tez_to_mutez (tz_minting p)))
+    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (Ligo.int (tz_minting p)))
     (Ligo.mul_int_int (fixedpoint_to_raw fixedpoint_one) (Ligo.int_from_literal "1_000_000"))
 
 (** Current liquidation price (in tez/kit). *)
 let liquidation_price (p: parameters) : ratio =
   make_ratio
-    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (tez_to_mutez (tz_liquidation p)))
+    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (Ligo.int (tz_liquidation p)))
     (Ligo.mul_int_int (fixedpoint_to_raw fixedpoint_one) (Ligo.int_from_literal "1_000_000"))
 
 (** Given the amount of kit necessary to close all existing burrows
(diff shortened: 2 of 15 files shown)


src/parameters.ml
```diff
@@ -8,8 +8,8 @@ open Constants
 
 type parameters =
   { q : fixedpoint; (* 1/kit, really *)
-    index: Ligo.tez;
-    protected_index: Ligo.tez;
+    index: Ligo.nat;
```
dorranh, line 11: Perhaps we should add a note here that index has units `tez/kit` for clarity.

src/parameters.ml
```diff
@@ -39,21 +39,21 @@ let initial_parameters : parameters =
...
   make_ratio
-    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (tez_to_mutez (tz_minting p)))
+    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (Ligo.int (tz_minting p)))
     (Ligo.mul_int_int (fixedpoint_to_raw fixedpoint_one) (Ligo.int_from_literal "1_000_000"))
 
 (** Current liquidation price (in tez/kit). *)
 let liquidation_price (p: parameters) : ratio =
   make_ratio
-    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (tez_to_mutez (tz_liquidation p)))
+    (Ligo.mul_int_int (fixedpoint_to_raw p.q) (Ligo.int (tz_liquidation p)))
```
dorranh, line 56: It's nice that we don't have to do the extra conversion from tez to int here :+1:

src/parameters.ml
```diff
@@ -246,10 +246,7 @@ let[@inline] compute_current_protected_index (last_protected_index: Ligo.tez) (c
           )
        )
     )
-    (Ligo.mul_int_int
-       protected_index_inverse_epsilon
-       (Ligo.int_from_literal "1_000_000")
-    )
+    protected_index_inverse_epsilon
```
dorranh, line 249: Where did the extra `1_000_000` in the denominator go here? Was that just required when using `fraction_to_tez_floor`?

src/checker.ml
```diff
@@ -834,7 +834,7 @@ let[@inline] touch_with_index (state: checker) (index:Ligo.tez) : (LigoOp.operat
 let entrypoint_touch (state, _: checker * unit) : (LigoOp.operation list * checker) =
   let index = match state.last_price with
     | None -> state.parameters.index (* use the old one *)
-    | Some i -> Ligo.mul_nat_tez i (Ligo.tez_from_literal "1mutez") in (* FIXME: Is the nat supposed to represent tez? *)
+    | Some i -> i in (* FIXME: Is the nat supposed to represent tez? *)
```
dorranh, line 837: Re: the FIXME. It seems that `state.last_price` is meant to have units tez/kit, right?

dorranh (commented):
Nice! Everything seems to be in order as far as I can tell. The only thing we might want to add would be the regression test we mentioned where we call touch `entrypoint_touch` with indices set sufficiently high for the old implementation to fail.

src/parameters.ml
gkaracha: Hmmm, actually, I am not entirely sure that these are the units for the index, now that you mention it. It was `tez` before, the [original doc](https://hackmd.io/teMO2x9PRRy1iTBtrSMBvA#Target-and-quantity) said _"expressed in kit"_, and now you mentioned `tez/kit` :joy: Agreed about figuring out (and adding comments about) the units of measure, but I think it might be best to do that in a separate PR.

src/parameters.ml
gkaracha: Yeah, that's nice indeed! I expect other places can be simplified too.

src/checker.ml
gkaracha: Well, I am still not entirely sure, that's why I've been reluctant to remove the `FIXME` here :thinking: I'll make a pass and see if I can (re-)figure out all the units of measure again in a separate PR, given how they are used.

src/parameters.ml
```diff
@@ -321,15 +318,15 @@ let[@inline] compute_current_q (last_q: fixedpoint) (last_drift: fixedpoint) (la
...
+let[@inline] compute_current_target (current_q: fixedpoint) (current_index: Ligo.nat) (current_kit_in_tez: ratio) : fixedpoint =
   let { num = num; den = den; } = current_kit_in_tez in
   fixedpoint_of_raw
     (fdiv_int_int
        (Ligo.mul_int_int
           den
           (Ligo.mul_int_int
              (fixedpoint_to_raw current_q)
-             (tez_to_mutez current_index)
+             (Ligo.int current_index)
```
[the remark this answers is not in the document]
gkaracha: Sweet! Yeah, I hadn't thought of that—will do :+1: In fact, I believe there might be other places where this can apply; I'll make a separate search after merging this PR for such cases. That'll probably reduce the byte count too :ok_hand:

tests/testArbitrary.ml
```diff
@@ -24,6 +24,12 @@ let arb_positive_lqt = QCheck.map (fun x -> lqt_of_denomination (Ligo.nat_from_l
 
 let arb_nat = QCheck.map (fun x -> Ligo.nat_from_literal ((string_of_int x) ^ "n")) QCheck.(0 -- max_int)
 
+(* somewhere between 0 and 3 *)
+let arb_small_nat =
+  QCheck.map
+    (fun x -> Ligo.nat_from_literal ((string_of_int x) ^ "n"))
+    QCheck.(1 -- ((max_int / 479_988_656_967) / 4))
```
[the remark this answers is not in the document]
gkaracha: Ah, I have no idea, I just made `arb_small_nat` as a copy of `arb_small_tez`, which was already convoluted (for no reason that I can see). I'll clean them both up :+1:

src/parameters.ml
gkaracha: Ah, that's a difference between `fraction_to_nat_floor` and `fraction_to_tez_floor`. `fraction_to_tez_floor` deals in `tez`, not `mutez`. So, for example
```ocaml
fraction_to_tez_floor (Ligo.int_from_literal "1_000_000") (Ligo.int_from_literal "1_000_000")
```
is equal to 1tez (which is 1_000_000mutez) and
```ocaml
fraction_to_tez_floor (Ligo.int_from_literal "1_234_567") (Ligo.int_from_literal "1_000_000")
```
is equal to `1_234_567mutez`, but
```ocaml
fraction_to_nat_floor (Ligo.int_from_literal "1_234_567") (Ligo.int_from_literal "1_000_000")
```
is simply equal to `1` (`= div 1234567 1000000`).

dorranh: Makes sense, thanks for the explanation!

src/parameters.ml
dorranh: :thinking: hmm good point. That approach also sounds good to me

src/parameters.ml
gkaracha: No prob! I do agree with @utdemir overall that scaling factors are a hassle and it would be nice to have less of them sprinkled over the codebase.

src/parameters.ml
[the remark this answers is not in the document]
gkaracha: Fixed this one in 89a5231, after we merge this PR I'll for more instances like this.

gkaracha:
@dorranh 
> The only thing we might want to add would be the regression test we mentioned where we call touch entrypoint_touch with indices set sufficiently high for the old implementation to fail.

Yeah, that'd be great indeed, though I am not sure about the exact conditions to trigger it. Did you manage to reproduce it using the e2e test?

tests/testArbitrary.ml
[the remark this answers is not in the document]
dorranh: I actually don't know anything more about this either :sweat_smile:. As far as I know there is nothing special about this number. In my case it looks like I've used this value for adding a random delta to a given amount of tez, but as far as I know slightly increasing or decreasing this value should make no difference in that context.

dorranh:
@gkaracha only roughly (not with an automated test). The conditions in the end to end test are essentially an index with initial oracle price `1_000_000` which gets increased to `100_000_000` and left for an arbitrary amount of time (10s of minutes). The only weird thing for us is until we change the representation of `Ligo.Tez` to `Int64`, I think we would need to somehow manually check for large tez values which are greater than the maximum `Int64`. Now that the index is of type `Nat`, I don't actually know if there is an easy way of checking this :thinking:

tests/testArbitrary.ml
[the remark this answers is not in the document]
gkaracha: Hahaha, nice. Now that I'm working on #192 I'll switch it to `10_000_000_000` then. I've had to change many of our `max_int` likewise anyway, since #192 uses a bounded representation for tez and otherwise tests would overflow on addition.

PR #192, opened 2021-07-12
Represent tez as an Int64.t

### LINKED ISSUE #190: Switch the internal representation of Ligo.tez to Int64.t
As noted by @dorranh, the [Michelson specification](https://gitlab.com/tezos/tezos/-/blob/master/docs/009/michelson.rst) clearly states that:

> _Mutez (micro-Tez) are internally represented by a 64 bit signed integers._

but our current mock implementation of tez in `ligo.ml` uses `Z.t` instead:
```ocaml
type tez = Z.t
```
We should probably change the definition of tez to the following:
```ocaml
type tez = Int64.t
```

gkaracha:
Fixes #190.

Many of our tests were dealing with very high amounts of tez, leading to failures after switching to `Int64.t`. So, I changed most generators for tez to go up to 10_000tez, to avoid hitting the `Int64.max_int` limit.


src/ligo.ml
```diff
@@ -240,46 +240,94 @@ let geq_timestamp_timestamp = Z.geq
...
+
+let mul_tez_nat x y =
+  assert (x >= Int64.zero);
+  assert (y >= Z.zero);
+  if x = Int64.zero then
+    Int64.zero
+  else if y > Z.of_int64 (Int64.div Int64.max_int x) then
+    failwith "Ligo.mul_tez_nat: overflow"
+  else
+    Int64.mul x (Z.to_int64 y)
```
gkaracha, line 291: Gotta hate GH sometimes :facepalm:

src/ligo.ml
```diff
@@ -240,46 +240,94 @@ let geq_timestamp_timestamp = Z.geq
...
-type tez = Z.t
+type tez = Int64.t
 
 let tez_from_literal s =
   try
     let n = parse_int_with_suffix "mutez" s in
     if Z.lt n Z.zero then
-      failwith "Ligo.tez_from_literal: negative"
-    else n
+      failwith "Ligo.tez_from_literal: underflow"
```
dorranh, line 249: I'm not sure if "underflow" or simply "negative" as it was before would make more sense in this message :thinking:

src/ligo.ml
```diff
@@ -240,46 +240,94 @@ let geq_timestamp_timestamp = Z.geq
...
-
-let mul_tez_nat = Z.mul
+    Int64.sub x y
+
+let mul_nat_tez x y =
+  assert (x >= Z.zero);
+  assert (y >= Int64.zero);
+  if y = Int64.zero then
+    Int64.zero
+  else if x > Z.of_int64 (Int64.div Int64.max_int y) then
```
dorranh, line 278: Will this condition also work for cases where max_int cannot be divided evenly by y?

src/ligo.ml
```diff
@@ -240,46 +240,94 @@ let geq_timestamp_timestamp = Z.geq
...
+    failwith "Ligo.mul_nat_tez: overflow"
+  else
+    Int64.mul (Z.to_int64 x) y
+
+let mul_tez_nat x y =
+  assert (x >= Int64.zero);
+  assert (y >= Z.zero);
+  if x = Int64.zero then
+    Int64.zero
+  else if y > Z.of_int64 (Int64.div Int64.max_int x) then
```
dorranh, line 288: Same question as above for this if condition

src/ligo.ml
dorranh: After thinking a bit more about the semantics of Int64.div I think this might be okay as-is, but I still just wanted to double check.

dorranh (commented):
Awesome, thanks for putting this together @gkaracha! It looks good to me in general. I just had a couple of comments / questions. I think I would also personally feel better to have a second review (@utdemir) before merging given how core of a change this is.

src/ligo.ml
gkaracha: Yeah, since `Int64.div` floors and all numbers involved are non-negative, I believe this is the safest.

src/ligo.ml
gkaracha: Same here, since `Int64.div` floors and all numbers involved are non-negative, I believe this is the safest.

src/ligo.ml
gkaracha: Ah, true. Since this is creation perhaps "negative" is more appropriate. Good point! :+1:

gkaracha:
> I think I would also personally feel better to have a second review (@utdemir) before merging given how core of a change this is.

Totally agree @dorranh, it's already uncomfortable enough changing something so fundamental so late, the more pairs of eyes check it the better!

tests/testTez.ml
```diff
@@ -1,6 +1,10 @@
 open OUnit2
 open TestLib
 
+(* TODO: Would be nice to have randomized tests, and even check the bounds
+ * tightly. Perhaps we should consider doing something similar for the other
+ * types (Lqt, Kit, Ctez). *)
+
```
utdemir, line 7: I don't really know how to implement this, but checking those implementations against `ligo` would be pretty nice too. eg. whether running `mul_tez_nat` on Ocaml side always give some results as `*` in ligo (specialised to the same type).

tests/testAvlModel.ml
```diff
@@ -32,7 +32,7 @@ let addr_gen = QCheck.Gen.(
   )
 
 let tez_gen = QCheck.Gen.(
-    map (fun x -> Ligo.tez_from_literal ((string_of_int x) ^ "mutez")) (0 -- max_int)
+    map (fun x -> Ligo.tez_from_literal ((string_of_int x) ^ "mutez")) (0 -- 10_000_000_000)
```
utdemir, line 35: This is a bit interesting.

If we're talking about some actual `tez`, this change is good, because I don't think there ever will be `2^64 mutez`).

But since there are places we get a `tez` parameter from outside, where they can possibly pass a really large number to us (eg. `checker.withdraw_tez((2**64, 1))`). It is not hard to imagine a situation where we do some calculations on it without checking for the overflow, then checking the now incorrect result against some number, and allowing the operation in.

So I think this is not a good _default_ for `tez_gen`, because it might give a false sense of security. We can always have another generator like `safe_tez_gen` which generates more realistic looking numbers.

utdemir (commented):
Looks good to me too, just one comment.

tests/testAvlModel.ml
gkaracha: That's a great point! :ok_hand: I will go through all entrypoints that receive tez not necessarily reflecting real value and make sure we generate numbers in the entire spectrum for those.

However, we have several generators for tez already, depending on the usecase, and the `tez_gen` generator above is used _exclusively_ in `testAvlModel.ml` to represent actual collateral contained in slices. So, for this one I think we should avoid going up to `max_int` (the problem being of course that we add them up in the avl tree, and we get trivial overflows that are impossible in reality).

What do you think?

tests/testTez.ml
gkaracha: Ah, yeah, that'd be great indeed. If we could have checked against Michelson we'd have caught this misrepresentation issue ages ago too :sweat_smile: I'm not entirely sure either, but I'll leave the `TODO` comment there, until we can figure something out. Alternatively we could open an issue, though it doesn't feel very pressing at the moment I guess.

tests/testAvlModel.ml
utdemir: Oh yeah, I do agree with that; when testing AVL there's no need to have unrealistic liquidation slices. My suggestion was only to rename the existing one to `safe_tez_gen` or something, so we won't forget about this caveat in future :).

gkaracha: Ah, I see. Will do!

By the way, about

> But since there are places we get a `tez` parameter from outside, where they can possibly pass a really large number to us (eg. `checker.withdraw_tez((2**64, 1))`). It is not hard to imagine a situation where we do some calculations on it without checking for the overflow, then checking the now incorrect result against some number, and allowing the operation in.

I am thinking that is not possible for this to happen, since tezos fails on overflows (now we'll do too). Even if the user passes an unrealistic value and there is some overflow, the whole operation will fail instead. :thinking:

utdemir: Oh, that's a great default. Yeah, this changes things quite a bit, we might even disregard my previous comment then :). No need to be scared of overflows if they abort th execution.

gkaracha: Great, I'll go ahead and mark this as "resolved" then! I went ahead and improved the names and added a couple of comments in 7d7acc3 anyway; some extra clarity is always welcome :slightly_smiling_face:

src/ligo.ml
gkaracha: Changed them both since I like me some symmetry :smile::
```ocaml
    [..]
    if Z.lt n Z.zero then
      failwith "Ligo.tez_from_literal: out of range (negative)"
    else if Z.gt n (Z.of_int64 Int64.max_int) then
      failwith "Ligo.tez_from_literal: out of range (positive)"
    else
    [..]
```

Issue #189, opened 2021-07-09
Entrypoint calls in end-to-end tests


dorranh:
As of writing this (5d4f3cf), we have end-to-end tests calling many but not all of Checker's entrypoints. As a part of our efforts to expand the coverage of our test suites, it would be good to at least call the remaining endpoints as a part of our end-to-end tests.

**Deployment**
- [X] deployMetadata
- [X] deployFunction
- [X] sealContract

**FA2 interface:** (#196)
- [x] transfer
- [x] balance_of (#197)
- [x] update_operators

**Main endpoints:**

Burrows: (#196)
- [X] create_burrow
- [X] mint_kit
- [x] burn_kit
- [X] mark_for_liquidation
- [X]	deposit_tez
- [X]	withdraw_tez
- [X]	touch_burrow
- [x] activate_burrow
- [X]	deactivate_burrow
- [x] set_burrow_delegate

Liquidations: (#193)
- [x] liquidation_auction_claim_win
- [x] touch_liquidation_slices
- [X] liquidation_auction_place_bid
- [x] cancel_liquidation_slice

CFMM:
- [X]	buy_kit
- [X]	sell_kit
- [X]	remove_liquidity
- [X] add_liquidity

General:
- [X] touch
- [X] receive_price (internal, called from oracle)
- [X]	receive_slice_from_burrow (internal, called from burrow)

dorranh:
With #197 merged, all of the entrypoints are called in our end to end tests :tada:

Issue #150, opened 2021-06-08
Off-chain views relating to liquidation auctions


gkaracha:
Perhaps the Web UI will not offer functionality around liquidation auctions, but I think we could (and perhaps should) offer some off-chain views to assist users interacting with liquidation auctions. Let's discuss ideas about views that might be useful for users in this issue.

gkaracha:
A few ideas to start with:

* Current winning bid
* Minimal amount of kit needed to place a bid
* How much tez is currently being auctioned off
* Blocks/Seconds passed since the last bid and how many remain until the auction closes

gkaracha:
As far as I can tell, our current view exposes all basic information for interacting with the current liquidation auction:
```ocaml
val view_current_liquidation_auction_details: (unit * checker) -> view_current_liquidation_auction_details_result

type view_current_liquidation_auction_details_result =
  { auction_id: liquidation_auction_id
  ; collateral: tok
  ; minimum_bid: kit
  ; current_bid: bid option
  ; remaining_blocks: Ligo.int option
  ; remaining_seconds: Ligo.int option
  }
```
So, I think we can close this one.

The only thing that is missing now is a view for identifying slices whose liquidation can be cancelled (see [my comment above](https://github.com/tezos-checker/checker/issues/150#issuecomment-892524022)), but that is not for participants in liquidation auctions, but rather for burrow owners. Let's open a separate issue for that (still low priority).

PR #193, opened 2021-07-13
Call remaining liquidation auction entrypoints in end to end tests

dorranh:
Adds calls to the remaining liquidation auction entrypoints which were previously not called in our end to end tests (#189). To allow this to run in a reasonable amount of time, I had to update our e2e patch to modify the `max_bid_interval_in_blocks` and `max_bid_interval_in_seconds` to much smaller values. Otherwise the test would have to wait up to 20 minutes for the auction to complete.


e2e/main.py
```diff
@@ -294,22 +326,89 @@ def call_bulk(bulks, *, batch_size):
...
+            ](),
+        ):
+            # Only retrieve the first queued slice ptr for efficiency
+            break
+        # Deposit a ton of tez to this burrow to ensure that it is no longer liquidatable
+        call_endpoint(
+            checker,
+            "deposit_tez",
+            leaf["leaf"]["value"]["contents"]["burrow"][1],
+            amount=100_000_000_000,
```
gkaracha, line 362: Ha, that's a ton of tez indeed!

e2e/main.py
```diff
@@ -294,22 +326,89 @@ def call_bulk(bulks, *, batch_size):
...
+        for leaf_ptr, leaf in auction_avl_leaves(checker, current_auctions_ptr):
+            # Double check that we are only working with leaf ptrs
+            assert "leaf" in leaf
+            auctioned_slices.append(leaf_ptr)
+
+        # Note: batching here based on empirical estimates of how many slices we can
+        # touch in one go before hitting the gas limit.
+        for i in range(0, len(auctioned_slices), 10):
+            slices_to_cancel = auctioned_slices[i : i + 10]
+            call_endpoint(checker, "touch_liquidation_slices", slices_to_cancel)
```
gkaracha, line 407: I wonder if we'd gain anything if we touched the slices a little out of order here, to cover more execution paths. It's certainly not necessary; just a thought :slightly_smiling_face:

gkaracha (approved):
Nice, LGTM!

e2e/main.py
dorranh: :moneybag:

e2e/main.py
dorranh: That's a good point. I think that ties in well with the goal of adding some profiling to these tests, so I noted it in a new issue here: https://github.com/tezos-checker/checker/issues/194.

gkaracha: Nice! :+1:

utdemir:
I'm a bit concerned on us looking at the underlying storage to get some information in order to write this test case; I feel like it would be more realistic if we were only using the public interface (entrypoints and views) to achieve this. 

It probably is the case where we don't expose enough views to achieve this. I guess:

* The winner of an auction has to have a way to see/touch the slices for their lot.
* A burrow owner has to be able to see/touch the slices for their burrow.

There might be more than one possble interface to achieve the above.

I guess, since we are just trying to measure the gas costs it might be fine; but still if some of the things we are doing are natural things to do on checker, we should have a nice way of doing those without having to understand the internal storage.

dorranh:
@utdemir, I totally agree - it was a bit unfortunate having to write a traversal to be able to touch all of the slices. I think that @gkaracha created an issue related to this (https://github.com/tezos-checker/checker/issues/150). Perhaps we can centralize ideas for additional views there?

gkaracha:
Ah, yes, that's a very good point @utdemir. Ideally that'd be the case for all our scenaria in the e2e tests I guess.

However: If the operations are just lookups in our storage we could offer them as views, but if they are calculations that we otherwise do not use elsewhere within checker it might not be a good idea to make views for those. They could get out of sync (and I expect advanced users will anyway look into the storage).

Either way I agree that #150 is a good place to keep all ideas for views, we should probably add the two you recommended to the list :+1:

PR #195, opened 2021-07-13
Override "burrow" in checker.ml to avoid using outdated values

gkaracha:
I was trying to transcribe our e2e test into a unit test to trigger
#191 when I came across a bug in `Checker.mark_for_liquidation`.
When calling `compute_min_kit_for_unwarranted` the input, untouched
burrow (`burrow`) was used, instead of the most recent, touched
burrow (`updated_burrow`).

This was easy to fix, but it revealed two issues:
* In `checker.ml` when updating burrows we'd bind the new burrow
  to `updated_burrow` in a few places, which allowed for this sort
  of situation. Easy to fix: in this PR I took the liberty of
  changing this pattern and shadow `burrow` so that there's always
  only one burrow available.

* This whole issue showed up when I got an assertion failure while
  transcribing our e2e test as a unit test.   Interestingly, the e2e test
  worked just fine, even if our   assertion was violated, since we
  erase our assertions before generating the LIGO code. I am
  thinking that perhaps we need a "testing deployment mode" which
  allows us to keep most (all?) of our assertions.

### DIFF
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -162,8 +162,8 @@ let entrypoint_touch_burrow (state, burrow_id: checker * burrow_id) : LigoOp.ope
   assert_checker_invariants state;
   let _ = ensure_no_tez_given () in
   let burrow = find_burrow state.burrows burrow_id in
-  let updated_burrow = burrow_touch state.parameters burrow in
-  let state = {state with burrows = Ligo.Big_map.update burrow_id (Some updated_burrow) state.burrows} in
+  let burrow = burrow_touch state.parameters burrow in
+  let state = {state with burrows = Ligo.Big_map.update burrow_id (Some burrow) state.burrows} in
   assert_checker_invariants state;
   (([]: LigoOp.operation list), state)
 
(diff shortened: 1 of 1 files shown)


utdemir (approved):
>  in this PR I took the liberty of changing this pattern and shadow burrow so that there's always only one burrow available.

I quite like this pattern too, I can even call this non-recursive let functionality that allows shadowing is one of the better choices of Ocaml over Haskell.

> I am thinking that perhaps we need a "testing deployment mode" which allows us to keep most (all?) of our assertions.

I agree that it might be handy. However that does increase the complexity, since we probably need to keep the existing "pretty-much-production-except-a-few-constants checker" too.

To me, it still feels like assertions and end to end tests belong in slightly different spaces. To me, the point of E2E tests is to check the final product on its natural environment. Even though it seems like having assertions should not change anything important, it's one more divergence from production (on top of the protected index/auction time hacks we already have). If we end up having it, then we'd probably need to have another "real E2E tests" test suite too, testing the production version (and things like gas costs).

There is another possibility I can think of; where we compile something similar to `assert_checker_invariants` to Michelson, and run it a few times off-chain on E2E tests. At least it wouldn't change the deployed contract. This also introduces complexity, so still not a good way.

I don't really know what's the best way to solve this to be honest. I guess rather than moving E2E tests towerds unit tests, moving our Ocaml tests towards E2E tests might make sense - eg. simulating more of the blockchain with multiple contracts, transfers, balances.

gkaracha:
> I quite like this pattern too, I can even call this non-recursive let functionality that allows shadowing is one of the better choices of Ocaml over Haskell.

After relying on it so much in the last months, I can't agree more, it's one of the features of OCaml I've been enjoying a lot as well.

> I don't really know what's the best way to solve this to be honest. I guess rather than moving E2E tests towerds unit tests, moving our Ocaml tests towards E2E tests might make sense - eg. simulating more of the blockchain with multiple contracts, transfers, balances.

Yes, I haven't been able to figure out a satisfactory way to deal with this problem either, but I agree that bringing the OCaml tests closer to the E2E tests is probably our best choice at the moment. I am not sure how effectively could we mock the rest of the blockchain infrastructure, but I certainly want the gap to get narrower. There are still quite a few code paths that we currently test only via e2e tests and not via unit/property-based tests (such as the one this PR addresses).

PR #197, opened 2021-07-14
Add call to FA2 balance_of entrypoint in end-to-end tests

dorranh:
Added a call to balance_of in the end-to-end test suite. Since
balance_of generally requires a contract to accept its callback,
this test uses pytezos' callback_view functionality which locally
interprets the entrypoint call and returns the emitted callback
operation. As such, this test doesn't provide an estimate of the
gas cost of balance_of but still provides a way to check 1.) whether
the entrypoint can successfully be called and 2.) the accurate of the
balance in the emitted callback operation.

### DIFF
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -227,9 +227,22 @@ class E2ETest(SandboxedTestCase):
         call_checker_endpoint("transfer", [fa2_transfer])
         assert_kit_balance(checker, checker_alice.key.public_key_hash(), 10)
 
-        # TODO: Not calling `balance_of` entrypoint in these tests at the moment since it
-        # would require a separate contract. If this is determined to be important we
-        # can add a simple contract for receiving balances and use it in these tests.
+        # `balance_of` requires a contract callback when executing on-chain. To make tests
+        # more light-weight and avoid needing an additional mock contract, we call it as a view.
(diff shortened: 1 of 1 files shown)


e2e/main.py
```diff
@@ -227,9 +227,22 @@ def call_checker_endpoint(
...
+            "requests": [
+                {"owner": checker_alice.key.public_key_hash(), "token_id": kit_token_id}
+            ],
+            "callback": None,
+        }
+        print(f"Calling balance_of as an off-chain view with {fa2_balance_of}")
+        balance = checker.balance_of(**fa2_balance_of).callback_view()
+        # Check that this balance agrees with the kit balance view
+        assert_kit_balance(
+            checker, checker_alice.key.public_key_hash(), balance[0]["nat_2"]
```
gkaracha, line 244: Ha, now there's a `nat_2`? We have to figure this out at some point :sweat_smile:

gkaracha (approved):

gkaracha:
LGTM! :ok_hand:

e2e/main.py
dorranh: Indeed, it's quite confusing :sweat_smile:

PR #198, opened 2021-07-15
Restore correct behavior for dune build

gkaracha:
Following a334f05 `dune build @install` would not work correctly,
because `src/dune` did not specify a public name (so "checker" was
assumed), but the opam file in src was named "src.opam" so the
names did not match and `dune build @install` would silently do
nothing.

To address this, this PR adds the public name "checker" to `dune/src`
and the accompanying (currently empty) opam file `checker.opam` at
the top-level.


utdemir (approved):
I'm happy if the tests pass :).

I'm just curious, what does `dune build @install` do?

gkaracha:
Thanks for having a look! :pray: 

> I'm just curious, what does `dune build @install` do?

It's supposed to build all public artifacts (i.e., the ones that would be installed if someone used checker as a library). See [here](https://dune.readthedocs.io/en/stable/usage.html#built-in-aliases).

PR #200, opened 2021-07-16
ci: Add 'mutate' cronjob

utdemir:
This PR adds a GitHub Actions workflow called "mutate", which runs our mutation script twice a day. See this page for an example run: https://github.com/tezos-checker/checker/runs/3084729232

It runs four tasks in parallel, each taking around 2 hours. It's not blocking anything, so I think it is okay for it to take this long. We can always reduce the `num-mutations`, or add more parallelization (up to 20) if it starts to bother us.

The changes include:

* A change to `mutate.py` which now requires the `src/` prefix before the arguments. I found it handy, since with it you can do shell expansions, eg: `src/{burrow,cfmm,constants}.ml` and get tab completion on the shell.
* Removal of some mutations from `mutate.py` that correspond to functions we do not have.

### DIFF
--- a/scripts/mutate.py
+++ b/scripts/mutate.py
@@ -71,7 +71,7 @@ MUTATION_GROUPS = [
         "leq_ratio_ratio",
         "geq_ratio_ratio",
         "gt_ratio_ratio",
-        # there is no eq_ratio_ratio
+        "eq_ratio_ratio",
     },
     # BOOL
     {"true", "false"},
--- a/scripts/mutate.py
+++ b/scripts/mutate.py
@@ -49,8 +49,8 @@ MUTATION_GROUPS = [
     {"mul_int_int", "sub_int_int", "add_int_int", "div_int_int"},
     {"eq_int_int", "lt_int_int", "gt_int_int", "leq_int_int", "geq_int_int"},
     # NAT
-    {"mul_nat_int", "sub_nat_int", "add_nat_int", "div_nat_int"},
-    {"eq_nat_int", "lt_nat_int", "gt_nat_int", "leq_nat_int", "geq_nat_int"},
+    # {"mul_nat_int"}, # note: there is no sub_nat_int add_nat_int", "div_nat_int"
+    # {"eq_nat_int", "lt_nat_int", "gt_nat_int", "leq_nat_int", "geq_nat_int"},
     # TEZ
     {"sub_tez_tez", "add_tez_tez"},
     {"eq_tez_tez", "lt_tez_tez", "gt_tez_tez", "leq_tez_tez", "geq_tez_tez"},
@@ -71,7 +71,7 @@ MUTATION_GROUPS = [
         "leq_ratio_ratio",
         "geq_ratio_ratio",
         "gt_ratio_ratio",
-        "eq_ratio_ratio",
+        # there is no eq_ratio_ratio
     },
     # BOOL
     {"true", "false"},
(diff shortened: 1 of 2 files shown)


scripts/mutate.py
```diff
@@ -71,7 +71,7 @@ class MutationType(Enum):
         "leq_ratio_ratio",
         "geq_ratio_ratio",
         "gt_ratio_ratio",
-        "eq_ratio_ratio",
+        # there is no eq_ratio_ratio
```
gkaracha, line 74: Actually there is `eq_ratio_ratio`, but it's only useful for the tests so indeed there's no reason to consider it at all when mutating :+1: In fact, now that I look at it, this is true for all the comparing functions in `ratio.ml`: `leq_ratio_ratio`, `geq_ratio_ratio`, `lt_ratio_ratio`, `gt_ratio_ratio`, and `eq_ratio_ratio`; I don't think there's any merit in trying to mutate any of them.

gkaracha (commented):
Nice work! I wonder though if adding this job would deplete our CI time. Do you know what is our current monthly limit? Also, I am thinking that it might be a good idea to store the report as an artifact, but I guess we could change that later.

scripts/mutate.py
utdemir: Ooh, right. I think I was grepping them under the `src/` so I thought they were unused. You are right about no point testing them, but maybe it would be useful as mutation targets. And at worst, they will be unused.

I will definitely fix the comment, but I think I should keep the `eq_ratio_ratio` there.

gkaracha: Yeah, fair point, you can leave all of them there, just leave a comment that we don't really expect those to fire :+1:

utdemir:
>  I wonder though if adding this job would deplete our CI time. Do you know what is our current monthly limit?

[As far as I can see](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits), total time is unlimited for open public projects. There are some limits for maximum number of parallel jobs or maximum time for a single run; but we are well below those limits.

> Also, I am thinking that it might be a good idea to store the report as an artifact, but I guess we could change that later.

Yes, that was initially my intention too. But in order to do that, it'd be useful to modify the Python script to print out a JSON instead (optionally hidden behind a flag), so it is computer-readable. I actually have a half written implementation on a git stash somewhere. I just didn't want the change to get too big.

gkaracha:
> > I wonder though if adding this job would deplete our CI time. Do you know what is our current monthly limit?
> 
> [As far as I can see](https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits), total time is unlimited for open public projects. There are some limits for maximum number of parallel jobs or maximum time for a single run; but we are well below those limits.

Awesome, that's what I understand as well from the page you linked. If CI passes you have the green light from me then!

> > Also, I am thinking that it might be a good idea to store the report as an artifact, but I guess we could change that later.
> 
> Yes, that was initially my intention too. But before we do that, I think we should modify the Python script to print out a JSON instead (optionally hidden behind a flag), so it is computer-readable. I actually have a half written implementation on a git stash somewhere. I just didn't want the change to get too big.

Good point :+1: Let's leave this change for another PR then.

gkaracha (approved):

scripts/mutate.py
```diff
@@ -72,6 +72,8 @@ class MutationType(Enum):
         "geq_ratio_ratio",
         "gt_ratio_ratio",
         "eq_ratio_ratio",
+        # most (all?) of these only appear on tests, but they are here for
+        # completeness' sake.
```
gkaracha, line 76: Yeah, certainly all, since they are between `BEGIN_OCAML`/`END_OCAML`. In fact, most of `ratio.ml` is now in this state; at the end I might just integrate the parts of `ratio.ml` we need into `common.ml` and the parts that are test-only into test modules :sweat_smile: This is low priority though I guess.

PR #202, opened 2021-07-20
Fix issue with units in initial auction bids

dorranh:
Fixes a bug in `start_liquidation_auction_if_possible` where the initial bid for a liquidation auction was calculated assuming the liquidation price was in units of `kit/tez` while it actually has units of `tez/kit`. Also added a unit / regression test which catches the error in the previous implementation.


gkaracha (approved):
Nice, thank you so much for fixing this! :pray: LGTM, but let's wait for the CI to finish before merging.

dorranh:
No worries, that sounds good :slightly_smiling_face:

utdemir:
Awesome find, thanks!

Issue #156, opened 2021-06-09
Ensure parameters.oustanding_kit does not drift far from the real value


gkaracha:
In the system parameters we keep track of the total amount of kit needed to close all currently open burrows via field `outstanding_kit`. However, this is only an approximation of the real amount. Not only we update the `burrow.outstanding_kit` field individually (per burrow) when touching burrows, but also burrows or checker might remain untouched for a while, which could introduce further rounding errors.

**Errors of a few percents per year are NOT acceptable. Errors of 0.1% or so per year would be tolerable.**

In order to test this meaningfully we should be able to create scenaria with many burrows, a lot of activity, and let a lot of time pass, which can be quite complicated to manufacture. Furthermore, one must keep track of all the burrows ever created to be able to inspect the storage and aggregate the total amount of outstanding kit for all burrows so that it can be compared to `parameters.oustanding_kit`.

PR #204, opened 2021-07-21
Expose the stats with a script and via a PR comment

utdemir:
This PR adds:

* A command line script, called `artifacts.py` that is intended to be a toolkit for Checker artifacts. It currently has two functionality:

  - plot-stats: Creates a plot of gas costs and entrypoint size changes up until the given commit.
  - compare-stats: Prints out the difference in terms of gas costs and entrypint sizes between two commits.

* A GitHub Action that is triggered by `pull_request` events, which runs the above `compare-stats` subcommand against the PR base and head, and comments the result.

There are that can be improved:

* There are more information that we can collect. Like coverage, or the accuracy of some checker approximations (issue #156).
* `plot-stats` command is not super useful; since the gas cost of `touch` dwarfs the others, and makes it hard to see the differences.
* The CI posts a new comment for each push. This might spam the thread if one keeps pushing new commits. If this starts bothering us, we might need to find another way of reporting the results.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,8 +35,9 @@ jobs:
         run: |
           cat result/functions.json \
             | jq --sort-keys '.lazy_functions | map({ key: .name, value: .chunks|add|length|(./2) }) | from_entries' \
-            > entrypoint-sizes.json
+            | tee entrypoint-sizes.json
       - uses: actions/upload-artifact@v2.2.4
+        if: github.event_name == 'push'
         with:
           name: stats
           path: entrypoint-sizes.json
@@ -61,12 +62,42 @@ jobs:
       - name: Generate profile plots
         run: nix-shell --run "python e2e/plot-gas-profiles.py gas_profiles.json --output auction-gas-profiles.png"
       - uses: actions/upload-artifact@v2.2.4
+        if: github.event_name == 'push'
         with:
           name: stats
           path: gas-costs.json
           if-no-files-found: error
       - uses: actions/upload-artifact@v2.2.4
+        if: github.event_name == 'push'
         with:
           name: gas-profile-plots
           path: auction-gas-profiles.png
           if-no-files-found: error
+  comment:
+    needs: [ build, e2e ]
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - uses: cachix/install-nix-action@v13
+      - uses: cachix/cachix-action@v10
+        with:
+          name: tezos-checker
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - uses: actions/checkout@v2
+      - name: Build dependencies
+        run: nix-shell --run ':'
+      - name: Calculate the diff
+        run: nix-shell --run './scripts/artifacts.py compare-stats --previous ${{ github.event.pull_request.base.sha }} --next ${{ github.event.pull_request.head.sha }} | tee msg.md'
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+      - uses: actions/github-script@v4.0.2
+        with:
+          script: |
+            const { promises: fs } = require("fs");
+            const msg = await fs.readFile("msg.md", { "encoding": "utf-8" });
+            github.issues.createComment({
+              issue_number: context.issue.number,
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              body: msg
+            })
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,8 @@ jobs:
       - name: Generate profile plots
         run: nix-shell --run "python e2e/plot-gas-profiles.py gas_profiles.json --output auction-gas-profiles.png"
       - uses: actions/upload-artifact@v2.2.4
+        # we only upload artifacts on 'push' events, this is simply to have one set of artifacts
+        # per commit.
         if: github.event_name == 'push'
         with:
           name: stats
--- a/scripts/artifacts.py
+++ b/scripts/artifacts.py
@@ -219,7 +219,7 @@ def diff_change(diff):
     if diff.previous is None:
         return diff.next
     elif diff.next is None:
-        return diff.previous
+        return -diff.previous
     else:
         return diff.next - diff.previous
 
--- a/scripts/artifacts.py
+++ b/scripts/artifacts.py
@@ -251,5 +251,3 @@ def print_diffs(diffs, *, title, header_previous, header_next):
 
 if __name__ == "__main__":
     cli()
-
-# import code; code.interact(local=locals())
--- a/scripts/artifacts.py
+++ b/scripts/artifacts.py
@@ -215,14 +215,6 @@ def get_first_line(s):
     return ls[0] if ls else ""
 
 
-def short_rev(rev):
-    return rev[:8]
-
-
-def list_getdefault(l, ix, d):
-    return l[ix] if ix < len(l) else d
-
-
 def diff_change(diff):
     if diff.previous is None:
         return diff.next
--- a/scripts/artifacts.py
+++ b/scripts/artifacts.py
@@ -1,0 +1,263 @@
+#!/usr/bin/env python
+
+import io
+import sys
+import json
+import yaml
+import zipfile
+import os.path
+import itertools as it
+from pprint import pprint
+from collections import namedtuple
+
+import click
+import requests
+from tqdm import tqdm
+import matplotlib.pyplot as plt
+
+ENDPOINT_GRAPHQL = f"https://api.github.com/graphql"
+ENDPOINT_REST = f"https://api.github.com"
+
+CommitInfo = namedtuple("CommitInfo", ["message", "ref", "rev", "workflow_ids"])
+CommitStats = namedtuple("CommitStats", ["info", "gas_costs", "entrypoint_sizes"])
+Diff = namedtuple("Diff", ["key", "previous", "next"])
+
+accessToken = os.getenv("GITHUB_TOKEN")
+if not accessToken:
+    with open(os.path.expanduser("~/.config/hub")) as f:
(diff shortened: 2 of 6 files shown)


utdemir:
The script only prints out the changes. It seems like the gas cost of `buy_kit` and `touch` might change across runs, so likely we'll see them often. I guess, alternatively we can set a thresholds, but arbitrary numbers are no fun.

scripts/artifacts.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+def short_rev(rev):
+    return rev[:8]
+
+
+def list_getdefault(l, ix, d):
+    return l[ix] if ix < len(l) else d
+
+
```
gkaracha, line 225: As far as I can tell neither `short_rev` nor `list_getdefault` is called anywhere, so we could probably drop them.

scripts/artifacts.py
```diff
@@ -0,0 +1,263 @@
...
+def fetch_commit_info(expr):
+    return fetch_commit_infos(expr, limit=1)[0]
+
+
+def mk_commit_stats(info):
+    stats = download_stats(info.workflow_ids)
+
+    gas_costs = stats.get("gas-costs", [{}])[0]
+    # TODO: Gas costs are strings for no reason.
+    gas_costs = {k: int(v) for k, v in gas_costs.items()}
```
gkaracha, line 175: True :thinking: Hmmm, from [what I can see](https://pytezos.org/quick_start.html?highlight=gas_limit#reveal-public-key) representing all items as strings is a PyTezos choice (including `gas_limit`). We could change our own generator to convert it to an int before we store the information in the file ([here](https://github.com/tezos-checker/checker/blob/4d1e384af141f86ce7957342a657320d2b430935/e2e/main.py#L266)) but then the past gas-cost files would become ill-formed. I'd probably leave it as is for now

.github/workflows/ci.yml
```diff
@@ -61,12 +62,42 @@ jobs:
       - name: Generate profile plots
         run: nix-shell --run "python e2e/plot-gas-profiles.py gas_profiles.json --output auction-gas-profiles.png"
       - uses: actions/upload-artifact@v2.2.4
+        if: github.event_name == 'push'
```
dorranh, line 65: Is the reason for the conditional here to avoid the issues with implicit commits for PRs that you mentioned on Slack? If so, could we add a comment somewhere in the CI file noting this to help remind us in the future?

scripts/artifacts.py
```diff
@@ -0,0 +1,263 @@
...
+                      }
+                    }
+                  }
+                }
+              }
+            }
+          }
+        }
+      }
+    }
```
dorranh, line 134: Some serious brackets here :sweat_smile:

scripts/artifacts.py
```diff
@@ -0,0 +1,263 @@
...
+
+def fetch_commit_info(expr):
+    return fetch_commit_infos(expr, limit=1)[0]
+
+
+def mk_commit_stats(info):
+    stats = download_stats(info.workflow_ids)
+
+    gas_costs = stats.get("gas-costs", [{}])[0]
+    # TODO: Gas costs are strings for no reason.
```
dorranh, line 174: I think that this is just a quirk of the relevant RPC response on tezos. Perhaps we can also open an issue to make sure that we are converting these to ints in the end-to-end tests?

scripts/artifacts.py
```diff
@@ -0,0 +1,263 @@
...
+            print(
+                f"| {diff.key.ljust(title_len)} | {str(diff.previous): <10} | {str(diff.next): <10} | {diff_change(diff): <10} |"
+            )
+        print()
+
+
+if __name__ == "__main__":
+    cli()
+
+# import code; code.interact(local=locals())
```
dorranh, line 263:
```suggestion
```

dorranh (approved):
Nice! Just a couple of small comments, but otherwise it looks good to me. Thanks for looking into this - it will be great to have the stats reported in an convenient way.

scripts/artifacts.py
```diff
@@ -0,0 +1,263 @@
...
+
+def list_getdefault(l, ix, d):
+    return l[ix] if ix < len(l) else d
+
+
+def diff_change(diff):
+    if diff.previous is None:
+        return diff.next
+    elif diff.next is None:
+        return diff.previous
```
gkaracha, line 230: Nitpick: I guess for consistency this should be negative:
```suggestion
        return (- diff.previous)
```
I wonder also what would happen if both were `None`. Perhaps we could first check for that:
```python
    if diff.previous is None and diff.next is None:
        return 0
    elif diff.previous is None:
        ...
```

gkaracha (approved):
Nice! I left a couple small comments, but overall it LGTM :slightly_smiling_face:

gkaracha:
> The script only prints out the changes. It seems like the gas cost of `buy_kit` and `touch` might change across runs, so likely we'll see them often. I guess, alternatively we can set a thresholds, but arbitrary numbers are no fun.

Yeah, no need for thresholds, I think. Being able to see how the numbers change should be enough :+1:

.github/workflows/ci.yml
utdemir: Yeah, this is something worth mentioning.

The main issue is that, using the Github GraphQL API, I couldn't find a way to easily figure out the events of the jobs that produced an artifact; so when the script is fetching the artifacts it fetches the artifacts both for the `push` event and for the `pull-request` event. These might not be the same, because the `pull-request` event tests the merged version of the PR. So, the easiest way for me to solve this was to not upload artifacts on `pull-request` events.

scripts/artifacts.py
utdemir: Yeah... A possible alternative is to use a GraphQL client library for Python that'd maybe allow us to form the query with first class values. But we only have a single query, so just hardcoding the request body also works just fine.

.github/workflows/ci.yml
```diff
@@ -35,8 +35,9 @@ jobs:
         run: |
           cat result/functions.json \
             | jq --sort-keys '.lazy_functions | map({ key: .name, value: .chunks|add|length|(./2) }) | from_entries' \
-            > entrypoint-sizes.json
+            | tee entrypoint-sizes.json
```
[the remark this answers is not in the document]
utdemir: Yeah, I also find having stuff in the logs handy :).

scripts/artifacts.py
utdemir: I think there's a way to do it gradually. If we change the E2E tests to emit `int`'s, then this cast will keep working. After a while, when `master` has the new version for a while, we can remove the now redundant `int` cast.

scripts/artifacts.py
utdemir: Yeah, I will just send a PR to cast them to integers on end-to-end tests, and we can fix this one after that one is merged to master.

scripts/artifacts.py
utdemir: They both can not be `None` (unfortunately, not by construction); since having `None` on one side means that the key is found in one direction but not on the other. When the key is not there at all, we won't have the `Diff` object. If this was Haskell, we would be using the `these` package :).

> Nitpick: I guess for consistency this should be negative:

That's right, thanks!

scripts/artifacts.py
dorranh: No complaints here, I just thought it was a bit funny :smile:

scripts/artifacts.py
gkaracha: Ah, right, great thinking! :ok_hand:

scripts/artifacts.py
gkaracha: Ah, I see. Fair enough

> If this was Haskell, we would be using the these package :).

:heart:

gkaracha:
> * There are more information that we can collect. Like coverage, or the accuracy of some checker approximations (issue [Ensure parameters.oustanding_kit does not drift far from the real value #156](https://github.com/tezos-checker/checker/issues/156)).

Certainly +1 for having the test coverage diff in the PR comment. Particularly before we start working on the generalization (#213), which I expect will change the coverage results quite some. (Might be a little annoying though, getting the percentages out of the `./_coverage/src/*.ml.html` files.)

utdemir:
> Might be a little annoying though, getting the percentages out of the ./_coverage/src/*.ml.html files.

There is a handy command line interface, so hopefully it shouldn't be very hard:

```
❯ dune build @run-fast-tests --instrument-with bisect_ppx --force 
...
❯ bisect-ppx-report summary
Coverage: 1859/2271 (81.86%)
❯ bisect-ppx-report summary --per-file
 27.16 %     85/313    src/avl.ml
 96.85 %    246/254    src/burrow.ml
  0.00 %      0/0      src/burrowTypes.ml
100.00 %    130/130    src/cfmm.ml
100.00 %      7/7      src/cfmmTypes.ml
 86.54 %    405/468    src/checker.ml
100.00 %    132/132    src/checkerEntrypoints.ml
 92.86 %     39/42     src/checkerMain.ml
100.00 %      2/2      src/checkerTypes.ml
100.00 %     89/89     src/common.ml
100.00 %     63/63     src/constants.ml
100.00 %     25/25     src/ctez.ml
100.00 %    101/101    src/error.ml
  0.00 %      0/0      src/fa12Interface.ml
100.00 %     64/64     src/fa2Interface.ml
100.00 %     19/19     src/fixedPoint.ml
100.00 %     38/38     src/kit.ml
 63.26 %    136/215    src/liquidationAuction.ml
  0.00 %      0/0      src/liquidationAuctionPrimitiveTypes.ml
100.00 %      1/1      src/liquidationAuctionTypes.ml
100.00 %     28/28     src/lqt.ml
100.00 %     16/16     src/mem.ml
100.00 %    195/195    src/parameters.ml
100.00 %      4/4      src/ptr.ml
 52.31 %     34/65     src/sliceList.ml
 81.86 %   1859/2271   Project coverage
 ```

gkaracha:
Ah, right, I didn't know about the `--per-file` but now that I see it, it totally makes sense :sweat_smile: Perhaps I should try and implement this one, to practice a little my CI skills :slightly_smiling_face:

PR #205, opened 2021-07-21
Move ratio.ml into the tests folder

gkaracha:
Since we stopped using the ratio abstraction throughout the codebase for efficiency (see #44, #46, #47, #50, #51, #52, and probably more), over 50% of `ratio.ml` has been used in tests only and we've had to deactivate coverage for it. This PR moves it to the test folder where it belongs, and keeps the commonly used (relevant) parts into `common.ml`.


utdemir (approved):
CI is happy so LGTM :). Always good to shrink the codebase.

gkaracha:
> Always good to shrink the codebase.

Totally agree, thanks for the review! :smile:

PR #206, opened 2021-07-21
Improvements to test coverage (89.56% => 90.05%)

gkaracha:
This is another attempt to improve test coverage. It increases coverage for the following files:

| source file          | from   | to
| ---------------------|--------|--------
| `src/common.ml`      | 97.75% | 100.00%
| `src/burrow.ml`      | 96.48% |  96.85%
| `src/checkerMain.ml` | 79.17% |  88.10%
| `src/checker.ml`     | 83.58% |  84.40%

There is still room for improvement (quite some, I think), but I've already seen that the following will not allow us to get to 100% coverage:

* aantron/bisect_ppx#377
* Can't get `LigoOp.Tezos.get_entrypoint_opt` to return `None` without emulating the chain somehow.
* Can't create meaningful values of type `Ligo.contract` to be able to call to `Balance_of` without emulating the chain somehow. This we still might be able to fake, but I am not sure how valuable that is.

### DIFF
--- a/src/burrow.ml
+++ b/src/burrow.ml
@@ -535,7 +535,12 @@ let burrow_request_liquidation (p: parameters) (b: burrow) : liquidation_result 
       let b_without_reward = { b with collateral = Ligo.sub_tez_tez (Ligo.sub_tez_tez b.collateral partial_reward) creation_deposit } in
       let tez_to_auction = compute_tez_to_auction p b_without_reward in
 
-      if (Ligo.lt_int_int tez_to_auction (Ligo.int_from_literal "0")) || (Ligo.gt_int_int tez_to_auction (tez_to_mutez b_without_reward.collateral)) then
+      (* FIXME: The property checked by the following assertion is quite
+       * intricate to prove. We probably should include the proof somewhere
+       * in the codebase. *)
+      assert (Ligo.gt_int_int tez_to_auction (Ligo.int_from_literal "0"));
+
+      if Ligo.gt_int_int tez_to_auction (tez_to_mutez b_without_reward.collateral) then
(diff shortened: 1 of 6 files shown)


src/burrow.ml
```diff
@@ -535,7 +535,12 @@ let burrow_request_liquidation (p: parameters) (b: burrow) : liquidation_result
       let b_without_reward = { b with collateral = Ligo.sub_tez_tez (Ligo.sub_tez_tez b.collateral partial_reward) creation_deposit } in
       let tez_to_auction = compute_tez_to_auction p b_without_reward in
 
-      if (Ligo.lt_int_int tez_to_auction (Ligo.int_from_literal "0")) || (Ligo.gt_int_int tez_to_auction (tez_to_mutez b_without_reward.collateral)) then
+      (* FIXME: The property checked by the following assertion is quite
```
dorranh, line 538: Definitely agree on this point. It could also save us some work in the future if we have to revisit it.

dorranh (approved):
Nice, looks good to me!

src/burrow.ml
gkaracha: Yeah, I think I'll typeset the proof I wrote and add it in comments or somewhere in the spec (in the [operational descriptions](https://checker.readthedocs.io/en/latest/operational.html) maybe?). I'll look into it tomorrow.

gkaracha:
Thanks for the review @dorranh! I'll go ahead and merge this one. The next module to increase test coverage for in my list is `src/checkerEntrypoints.ml`, currently being at 58.33%. That one needs a lot of tests so I'll do that in a separate PR.

PR #207, opened 2021-07-22
Improvements to test coverage, reloaded (90.05% => 93.00%)

gkaracha:
This adds another set of tests to improve test coverage. It increases coverage for the following files:

| source file          | from   | to
| ---------------------|--------|--------
| `src/checkerEntrypoints.ml`      | 58.33% | 100.00%
| `src/checkerMain.ml`      | 88.10% |  92.86%

@utdemir @dorranh I might have stumbled upon a bug (or not, I still have to investigate the `FIXME` I left in the comments), but I wouldn't investigate it as part of this PR I think.

### DIFF
--- a/tests/testCheckerMain.ml
+++ b/tests/testCheckerMain.ml
@@ -12,6 +12,11 @@ open LiquidationAuctionPrimitiveTypes
 let property_test_count = 100
 let qcheck_to_ounit t = OUnit.ounit2_of_ounit1 @@ QCheck_ounit.to_ounit_test t
 
+(* This function lets us test all different branches of
+ * CheckerEntrypoints.lazyParamsToLazyFunctionId. This is the reason we pass
+ * the lazy parameters here, instead of the function id and the bytes
+ * separately, so that we can call lazyParamsToLazyFunctionId on each possible
+ * argument and increase test coverage. *)
 let test_deploy_function_with_lazy_params_succeeds lazy_params =
   Ligo.Tezos.reset ();
   Ligo.Tezos.new_transaction ~seconds_passed:0 ~blocks_passed:0 ~sender:leena_addr ~amount:(Ligo.tez_from_literal "0mutez");
--- a/tests/testCheckerMain.ml
+++ b/tests/testCheckerMain.ml
@@ -2,6 +2,7 @@ open OUnit2
 open TestLib
 open CheckerMain
 open Error
+open LiquidationAuctionPrimitiveTypes
 
 (* NOTE: The tests in this file mainly try to cover all execution paths in
  * CheckerMain.main, but don't do much more. At this high a level most of the
@@ -10,6 +11,24 @@ open Error
 
 let property_test_count = 100
 let qcheck_to_ounit t = OUnit.ounit2_of_ounit1 @@ QCheck_ounit.to_ounit_test t
+
+let test_deploy_function_with_lazy_params_succeeds lazy_params =
+  Ligo.Tezos.reset ();
+  Ligo.Tezos.new_transaction ~seconds_passed:0 ~blocks_passed:0 ~sender:leena_addr ~amount:(Ligo.tez_from_literal "0mutez");
+
+  let wrapper = CheckerMain.initial_wrapper leena_addr in (* unsealed *)
+
+  let fn_id, fn_bytes = CheckerEntrypoints.(lazyParamsToLazyFunctionId lazy_params) in
+  let op = CheckerMain.DeployFunction (fn_id, fn_bytes) in
+
+  (* This call should succeed (first time, no previous entry). *)
+  let _ops, wrapper = CheckerMain.main (op, wrapper) in
+  (* This call should also succeed (second time, concatenation). *)
+  (* Note: it does not really make sense deploying the exact same bytes twice,
+   * but DeployFunction is not meant to inspect the bytes, only concatenate
+   * them, so the call should succeed regardless. *)
+  let _ops, _wrapper = CheckerMain.main (op, wrapper) in
+  ()
 
 let suite =
   "CheckerMainTests" >::: [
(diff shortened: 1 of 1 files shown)


tests/testCheckerMain.ml
```diff
@@ -11,6 +12,32 @@ open Error
 let property_test_count = 100
 let qcheck_to_ounit t = OUnit.ounit2_of_ounit1 @@ QCheck_ounit.to_ounit_test t
 
+let test_deploy_function_with_lazy_params_succeeds lazy_params =
+  Ligo.Tezos.reset ();
```
dorranh, line 16: I was a bit confused at first about why we had to pass lazy_params when calling DeployFunction instead of just the function bytes. I see that it's used for looking up the function id/bytes, but perhaps a comment here explaining this might help make this more clear.

tests/testCheckerMain.ml
```diff
@@ -113,7 +273,205 @@ let suite =
...
+          (* This call should succeed *)
+          Ligo.Tezos.new_transaction ~seconds_passed:0 ~blocks_passed:0 ~sender:alice_addr ~amount:(Ligo.tez_from_literal "0mutez");
+          let _ops, _wrapper = CheckerMain.main (op, sealed_wrapper) in
+          ()
+       )
+    );
+
+    (* Note: rather superficial call, to increase path coverage;
+     * I would have liked a call that succeeds here instead, but
+     * #150 makes this annoying to achieve. *)
```
dorranh, line 429: :+1: Adding more views would make the end-to-end tests less verbose as well.

tests/testCheckerMain.ml
```diff
@@ -113,7 +273,205 @@ let suite =
...
+          let op = CheckerMain.(CheckerEntrypoint (LazyParams (Touch ()))) in
+          let _ops, sealed_wrapper = CheckerMain.main (op, sealed_wrapper) in
+
+          (* Mark_for_liquidation *)
+          Ligo.Tezos.new_transaction ~seconds_passed:342 ~blocks_passed:5 ~sender:user_addr ~amount:(Ligo.tez_from_literal "0mutez"); (* the user themselves can mark it *)
+          let op = CheckerMain.(CheckerEntrypoint (LazyParams (Mark_for_liquidation (user_addr, burrow_id)))) in
+          let _ops, sealed_wrapper = CheckerMain.main (op, sealed_wrapper) in
+
+          (* Note: I would have liked to be able to recollateralize the burrow and try
+           * out Cancel_liquidation_slice here, but to do that we need to be able to find
```
dorranh, line 384: We can also potentially use some of the functionality from `sliceList.ml` to simplify this (e.g. `slice_list_from_auction_state`) though we actually haven't implemented any extra OCaml functions in that module for traversing the list :slightly_frowning_face:.

tests/testCheckerMain.ml
```diff
@@ -113,7 +273,205 @@ let suite =
...
+          (* Liquidation_auction_place_bid *)
+          let min_bid = CheckerEntrypoints.wrapper_view_current_liquidation_auction_minimum_bid ((), sealed_wrapper) in
+          Ligo.Tezos.new_transaction ~seconds_passed:394 ~blocks_passed:6 ~sender:user_addr ~amount:(Ligo.tez_from_literal "0mutez");
+          let op = CheckerMain.(CheckerEntrypoint (LazyParams (Liquidation_auction_place_bid (min_bid.auction_id, min_bid.minimum_bid)))) in
+          let _ops, sealed_wrapper = CheckerMain.main (op, sealed_wrapper) in
+
+(*
+          (* FIXME: WE HAVE A PROBLEM I THINK: THE FOLLOWING RAISES AN
+           * internalError_KitSubNegative error, which I thought was impossible.
+           * Is that because of how high the index is? Is there a bug somewhere? *)
```
dorranh, line 402: Ah that's too bad, but nice catch! Interested to see what causes this.

dorranh (approved):
Awesome, that's a big jump in coverage. Just one minor point about potentially adding a comment, but LGTM!

tests/testCheckerMain.ml
gkaracha: Ah, indeed. The reason I added all these tests was to increase test coverage for `lazyParamsToLazyFunctionId`. That's why these tests are in `testCheckerMain.ml`, but they actually increase test coverage for `checkerEntrypoints.ml`. I'll add a comment about it :+1:

tests/testCheckerMain.ml
gkaracha: Yeah, overall I'd like to follow the approach @utdemir had suggested: ideally we'd be able to have testing at such a high level using only the public API (entrypoints + offline views), and not inspecting the storage (much :upside_down_face:). This for both the unit tests at the level of `CheckerMain` and the e2e tests indeed.

tests/testCheckerMain.ml
gkaracha: This comment I left relates to the other one; ideally we would be able to figure out as burrow owners which slices of ours are in the queue in a more opaque way (offline view?). The scenario in the comment though is just a thought, I know we already test cancellations in the e2e tests (thanks about that btw :smile:) so I am not very worried about this one here. I do think though that we should look into #150 soon, not just for users, but also for our own convenience :sweat_smile:

tests/testCheckerMain.ml
gkaracha: Yeah, I am gonna look into this today, let's see what we find. I hope it is a bug in the implementation, but I gather it might be an issue with the spec as well. I believe that to place such a big bid, `user_addr` needs to utilize some of the kit obtained as a reward for touching checker a few calls above, but that was minted out of thin air (so it's part of the circulating kit, but not part of the outstanding kit). Touching the slices at this point then tries to burn more kit than was supposed to be burnable. I'll get to the bottom of this, but I'm afraid it might be a problem with the spec minting kit as part of calling `%touch` :confused:

gkaracha:
Thanks for the review @dorranh! :pray: I added a comment to clarify the use of `test_deploy_function_with_lazy_params_succeeds`; I'll merge once the CI is done :+1:

PR #208, opened 2021-07-23
Add extra unit tests for sliceList.ml

dorranh:
This PR increases test coverage for `sliceList.ml` to what should be 100%. Sadly the known issue with `bisect_ppx` and calling `Ligo.failwith` prevent the file from being reported as 100%.

### DIFF
--- a/src/sliceList.ml
+++ b/src/sliceList.ml
@@ -161,11 +161,17 @@ let[@inline] slice_list_remove (l:slice_list) (auctions:liquidation_auctions) (e
       else
         (* Case 2: We are removing the oldest slice *)
       if ptr = bounds.slice_list_oldest_ptr then
-        (* Case 2.1 it is the only element (i.e. also the youngest slice). The list is now empty *)
         match slice.younger with
-        | None ->
-          assert (ptr = bounds.slice_list_youngest_ptr);
-          (None: slice_list_bounds option)
+        (* Case 2.1 it is the only element (i.e. also the youngest slice). The list is now empty *)
+        (* NOTE: This branch is unreachable in the current implementation since slice lists of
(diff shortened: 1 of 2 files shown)


src/sliceList.ml
```diff
@@ -161,11 +161,17 @@ let[@inline] slice_list_remove (l:slice_list) (auctions:liquidation_auctions) (e
...
-        | None ->
-          assert (ptr = bounds.slice_list_youngest_ptr);
-          (None: slice_list_bounds option)
+        (* Case 2.1 it is the only element (i.e. also the youngest slice). The list is now empty *)
+        (* NOTE: This branch is unreachable in the current implementation since slice lists of
+         * a single element have slice_list_oldest_ptr = slice_list_youngest_ptr != None. This
+         * means that any single-element lists will go down the single element if- branch above
+         * (case 1.1), but we still need to perform pattern patching here since slice_list_bounds
+         * is of type option. Disabling coverage for this branch since it is unreachable.
+        *)
```
gkaracha, line 171: Woof, that's a risky territory; the `[@coverage off]` really deactivates coverage for the whole clause (including the pattern). :thinking: I agree with the logic, the branch must really be inaccessible, but perhaps it would be better to leave this clause be (that is, leave the very informative comment you added, but remove the `[@coverage off]`). We won't get to 100% anyway, so I'd feel safer if we didn't deactivate coverage here.

gkaracha (approved):
Nice! The tests look fine, but I'd say remove that `coverage off` annotation (see comment), just to be safe.

src/sliceList.ml
dorranh: True, I suppose that it makes sense to be a bit conservative on this front to avoid trouble in the future. Went ahead and re-enabled coverage here.

gkaracha: Thanks! :pray:

PR #210, opened 2021-07-26
Add mechanism for ignoring lines in mutation tests

dorranh:
Added a basic mechanism to `mutate.py` which allows us to skip lines containing an in-line `(* SKIP_MUTATION *)` comment. This uses the same rough approach that the script uses for assertions, which is faster than using a regex but will potentially catch stuff like string literals in the code with content like `let x = "SKIP_MUTATION_FOR_SOME_REASON" in ...`. There aren't any existing examples of this in the code base though, and I think we are fairly safe from running into such issues.


gkaracha (approved):
LGTM!

PR #211, opened 2021-07-26
Correct treatment of excess when burning kit

### LINKED ISSUE #136: Single source of truth for circulating kit and liquidity
To conform to [TZIP-012](https://gitlab.com/tzip/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#off-chain-views) we need a `total_supply` view which provides the total supply of kit and liquidity tokens. Currently the former is stored as part of the parameters (`parameters.circulating_kit`) and the latter as part of the cfmm state (`cfmm.lqt`). Furthermore, the kit owned by the cfmm is currently tracked in the cfmm (`cfmm.kit`), but not on the ledger. To avoid inconsistencies it would be better to have a single source of truth: the ledger in the `fa2_state`.

Related: #133.

### LINKED ISSUE #209: internalError_KitSubNegative error raised
While developing #207 I came across a scenario which raises an internal error (`internalError_KitSubNegative`). I tried to simplify the example a bit though it probably can be simplified further. The scenario goes as follows:

1. Initial state `(timestamp:0, level:0)`:
   ```
   last_index = 1_000_000n
   circulating = 0mukit
   outstanding = 0mukit
   ```

2. After `(~seconds_passed:0 ~blocks_passed:0)`, create a burrow with `10_000_000mutez` (1tez deposit + 9tez collateral).                 
   ```
   circulating = 0mukit
   outstanding = 0mukit
   ```

3. After `(~seconds_passed:1181 ~blocks_passed:18)`, mint as much kit as possible (`4_285_714mukit`).
   ```
   circulating = 4_285_714mukit
   outstanding = 4_285_714mukit
    alice_addr : 4_285_714mukit
   ```

4. After `(~seconds_passed:0 ~blocks_passed:0)`, set the last observed index to `1_357_906n`

5. After `(~seconds_passed:(60 * 191) ~blocks_passed:191)`, touch checker (last index still at `1_357_906n`).
   ```
   circulating = 205_969_055mukit  -- 4_285_714mukit minted + 201_683_333 touch reward + 8mukit cfmm accrual
   outstanding = 4_285_807mukit    -- 4_285_714mukit minted scaled up (burrowing fees and/or imbalance)
    alice_addr : 205_969_047mukit  -- 4_285_714mukit minted + 201_683_333 touch reward
    checker    : 8mukit            -- accrual of burrow fees to cfmm.kit
   ```

6. After `(~seconds_passed:342 ~blocks_passed:5)`, mark the burrow for liquidation.
   ```
   circulating = 205_969_055mukit  -- (as it was) 4_285_714mukit minted + 201_683_333 touch reward + 8mukit cfmm accrual
   outstanding = 4_285_807mukit    -- (as it was) 4_285_714mukit minted scaled up (burrowing fees and/or imbalance)
    alice_addr : 205_969_047mukit  -- (as it was) 4_285_714mukit minted + 201_683_333 touch reward
    checker    : 8mukit            -- (as it was) accrual of burrow fees to cfmm.kit
   ```

7. After `(~seconds_passed:63 ~blocks_passed:1)`, touch checker to start an auction (last index still at `1_357_906n`).
   ```
   circulating = 206_644_055mukit -- (+675_000mukit touch reward)
   outstanding = 4_285_809mukit   -- (+2mukit (burrowing fees and/or imbalance))
    alice_addr : 206_644_047mukit -- (+675_000mukit touch reward)
    checker    : 8mukit           -- (as it was) accrual of burrow fees to cfmm.kit
   ```

8. After `(~seconds_passed:394 ~blocks_passed:6)`, place a bid of `4_285_824mukit`.
   ```
   circulating = 206_644_055mukit -- (as it was)
   outstanding = 4_285_809mukit   -- (as it was)
    alice_addr : 202_358_223mukit -- (-4_285_824mukit auction bid)
    checker    : 4_285_832mukit   -- (8mukit in cfmm.kit + 4_285_824mukit auction bid)
   ```

9. If we touch checker now, what happens is that the received kit (the bid, equal to `4_285_824mukit`) is subtracted (burned) from the outstanding kit (currently at `4_285_820mukit`, `4mukit` less than the bid) and we get a failure.

Each individual step above makes sense to me, so I am not entirely sure how to classify this error. Some thoughts:
* On one hand, I am troubled that the out-of-thin-air reward can increase the kit in circulation an arbitrary amount (that would be a specification bug,I think).
* On the other hand, I assume that the total outstanding can always be decreased due to other reasons (e.g., if the value of kit rises); the implementation should be able to deal with this (that is, never try to burn more than the total outstanding kit) (that would be an implementation bug, I think).
* On the other-other hand (:upside_down_face:) I find it weird that we can have kit in circulation without it being backed by tez. I guess the touch reward is in a way justified/backed by the gas costs paid when touching?

Any thoughts?

gkaracha:
This PR fixes a long-standing bug around the concept of excess kit and removes the `excess_kit` field in burrows. It's still work in progress but I thought I'd open the PR early so that perhaps you could have a look too @utdemir @dorranh; the changes are rather pervasive (though necessary, I believe) and I really could use more pairs of eyes. Let me know what you think! :slightly_smiling_face: 

Closes #209.
Closes #136.

Summary:

- [X] The `excess_kit` is removed from the representation of burrows. This means that in cases where more than `burrow.outstanding_kit` is attempted to be burned/deposited, the excess is directly credited to the burrow owner on the FA2 ledger. This applies both to explicit burns (via `burrow_burn_kit`) and implicit burns (via `burrow_return_kit_from_auction`).
- [x] `assert_burrow_invariants` now has no invariants to check. I've left it for the time being, but I think we can just get rid of it.
- [X] Some information is now "leaking" from `burrow_burn_kit` and `burrow_return_kit_from_auction`: they both need to return to the caller how much of that kit really got burned and how much did not. This is a good thing. This change allows us to remove from `parameters.circulating_kit` and the `parameters.outstanding_kit` the kit that got repaid, **but not the part that did not**. This was the source of #209: previously we would erroneously try to remove all the kit we tried to burn from the `parameters.outstanding_kit`, even if some of it really ended up in the `excess_kit` field. I left a big comment about what happens with each part of the kit to be burned (and a few assertions) in `touch_liquidation_slice` to clarify this.
- [ ] There is still a hidden assumption in this logic: by never burning more kit than is outstanding _per burrow_, my text above implies that I don't expect there to be an underflow in the _global outstanding kit_ (`parameters.outstanding_kit`), which is what #209 is about. I have not proven this yet, but I believe that the calculations imply that `parameters.outstanding_kit >= sum burrow_i.outstanding_kit`. If that is indeed the case, then sound burning per burrow implies sound burning for the total outstanding kit. I expect us to observe this inequality when we address #156, but I will try to prove it before merging this PR anyway.
   If the inequality does not hold then we can always circumvent the issue of #209 by a simple if-then-else: never try to burn more kit from `parameters.outstanding_kit` than there is. I don't like this catch-all because it could hide other bugs, but if the inequality above really does not hold at least this if-then-else works as an absorbing bound at 0, essentially showing that we have to burn more kit than we thought were outstanding. If anything, it reduces the discrepancy between `parameters.outstanding_kit` and `sum burrow_i.outstanding_kit`.
- [X] I made `touch_liquidation_slice` and friends take bigger parts of the state as arguments to be able to update the FA2 ledger when we give kit to the burrow owners. In general I wrote things having clarity in mind, but we have to make sure that the gas costs don't increase too much because of this before merging (if so, we have to re-uglify the code). 
- [x] I have not looked at the docs yet, but I am sure I'd have to update them to reflect some of these changes.
- [x] The test folder still have debugging prints in it. I'll remove the calls when done (we can still keep the functions though; very handy).

### DIFF
--- a/src/burrow.ml
+++ b/src/burrow.ml
@@ -23,8 +23,6 @@ type burrow =
     collateral : Ligo.tez;
     (* Outstanding kit minted out of the burrow. *)
     outstanding_kit : kit;
-    (* Excess kit returned by auctions. *)
-    excess_kit : kit;
     (* The imbalance adjustment index observed the last time the burrow was
      * touched. *)
     adjustment_index : fixedpoint;
@@ -59,13 +57,10 @@ type liquidation_result = (liquidation_type * liquidation_details) option
 [@@@coverage on]
 
 let[@inline] assert_burrow_invariants (_b: burrow) : unit =
-  assert (_b.outstanding_kit = kit_zero || _b.excess_kit = kit_zero);
+  (* FIXME: no properties left to check *)
   ()
 
-(** Update the outstanding kit (and excess_kit), update the adjustment index,
-  * and the timestamp. When updating the outstanding kit we round up, to
-  * avoid someone touching their burrow all the time to prevent the fee from
-  * accumulating. *)
+(** Update the outstanding kit, update the adjustment index, and the timestamp. *)
 let burrow_touch (p: parameters) (burrow: burrow) : burrow =
   assert_burrow_invariants burrow;
   let burrow_out = if p.last_touched = burrow.last_touched
@@ -156,15 +151,6 @@ let burrow_max_mintable_kit (p: parameters) (b: burrow) : kit =
       (Ligo.mul_int_int num_fm num_mp) in
   kit_of_fraction_floor numerator denominator
 
-(** Rebalance the kit inside the burrow so that either outstanding_kit is zero
-  * or b.outstanding_kit is zero. *)
-let rebalance_kit (b: burrow) : burrow =
-  let kit_to_move = kit_min b.outstanding_kit b.excess_kit in
-  { b with
-    outstanding_kit = kit_sub b.outstanding_kit kit_to_move;
-    excess_kit = kit_sub b.excess_kit kit_to_move;
-  }
-
 let burrow_return_slice_from_auction
     (slice: liquidation_slice_contents)
     (burrow: burrow)
@@ -183,18 +169,22 @@ let burrow_return_slice_from_auction
 let burrow_return_kit_from_auction
     (slice: liquidation_slice_contents)
     (kit: kit)
-    (burrow: burrow) : burrow =
+    (burrow: burrow) : burrow * kit * kit =
   assert_burrow_invariants burrow;
   assert (Ligo.geq_tez_tez burrow.collateral_at_auction slice.tez);
+
+  let returned_kit = kit_min burrow.outstanding_kit kit in
+  let excess_kit = kit_sub kit returned_kit in
+
   let burrow_out =
-    rebalance_kit
-      { burrow with
-        excess_kit = kit_add burrow.excess_kit kit;
-        collateral_at_auction = Ligo.sub_tez_tez burrow.collateral_at_auction slice.tez;
-      }
-  in
+    { burrow with
+      outstanding_kit = kit_sub burrow.outstanding_kit returned_kit;
+      collateral_at_auction = Ligo.sub_tez_tez burrow.collateral_at_auction slice.tez;
+    } in
+
   assert (burrow.address = burrow_out.address);
-  burrow_out
+  assert (eq_kit_kit (kit_add returned_kit excess_kit) kit);
+  (burrow_out, returned_kit, excess_kit)
 
 let burrow_create (p: parameters) (addr: Ligo.address) (tez: Ligo.tez) (delegate_opt: Ligo.key_hash option) : burrow =
   if Ligo.lt_tez_tez tez creation_deposit
@@ -205,7 +195,6 @@ let burrow_create (p: parameters) (addr: Ligo.address) (tez: Ligo.tez) (delegate
       delegate = delegate_opt;
       collateral = Ligo.sub_tez_tez tez creation_deposit;
       outstanding_kit = kit_zero;
-      excess_kit = kit_zero;
       adjustment_index = compute_adjustment_index p;
       collateral_at_auction = Ligo.tez_from_literal "0mutez";
       last_touched = p.last_touched; (* NOTE: If checker is up-to-date, the timestamp should be _now_. *)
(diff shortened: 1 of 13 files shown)


src/burrow.ml
```diff
@@ -59,13 +57,10 @@ type liquidation_result = (liquidation_type * liquidation_details) option
 [@@@coverage on]
 
 let[@inline] assert_burrow_invariants (_b: burrow) : unit =
-  assert (_b.outstanding_kit = kit_zero || _b.excess_kit = kit_zero);
+  (* FIXME: no properties left to check *)
```
dorranh, line 60: :tada:

gkaracha:
So @utdemir @dorranh, this should be ready for reviewing (@dorranh not many things changed since the last time you had a look, but you might want to skim it). The size and gas cost differences I think are good; only `touch` and `touch_liquidation_slices` increased on both accounts, because of this change I mentioned:

> I made touch_liquidation_slice and friends take bigger parts of the state as arguments to be able to update the FA2 ledger when we give kit to the burrow owners. In general I wrote things having clarity in mind, but we have to make sure that the gas costs don't increase too much because of this before merging (if so, we have to re-uglify the code).

I think the clarity is worth the few extra bytes/gas, especially given that that's where the bug was before. Give me a shout if you disagree and I can try and change `touch_liquidation_slice`.

I started working on paper to prove my postulate that `parameters.outstanding_kit` is always greater-than-or-equal to the sum of all the outstanding kit in each burrow, but the code itself will remain as-is (I'll add a comment and assertions, once I finish the proof, but that's all).

tests/testCheckerMain.ml
```diff
@@ -43,6 +43,21 @@ let set_last_price_in_wrapper wrapper price_option =
       {wrapper with deployment_state = Sealed {state with last_price = price_option}}
   )
 
+let debug_print_all_kit_in_sealed_state msg wrapper =
```
dorranh, line 46: This might actually be something useful to drop into `testLib.ml` in case we need to do similar debugging in the future.

dorranh (approved):
Nice! I looked through a second time, and it seems to all be in order as far as I can tell. I am a bit surprised that we didn't need to change more tests for this instead of mostly needing to just remove obsolete ones. Though I suppose that it kind of makes sense since this change is largely internal and doesn't seem to affect a given account's balance in the fa2 ledger as far as I can tell.

gkaracha:
> I am a bit surprised that we didn't need to change more tests for this instead of mostly needing to just remove obsolete ones.

Right?! I thought so too and was not amused that not many tests failed. I think that the tests not failing means that our coverage could be improved. For example the changes to `burrow_burn_kit` reveal that no test we have tries to burn more kit than is outstanding. We might want to consider failing or crediting the rest of the kit back to the owner (as we currently) do, but the exact choice is not that important; the fact that no test triggers this behavior is :sweat_smile:

tests/testCheckerMain.ml
gkaracha: Ah, right, good point. Move it :+1:

PR #215, opened 2021-07-28
Add test coverage to the stats shown by the message bot

gkaracha:
This PR
* adds test coverage to our artifact (`test-coverage.json`), as reported by `bisect_ppx`
* extends the `compare-stats` subcommand to also print the test coverage difference between the two commits (usually base commit and HEAD).

I tried to reuse the infrastructure from #204 as much as possible, but there are a few differences between test coverage and the other two (gas costs and entrypoint sizes) which might be important enough to require us to treat test coverage differently:
1. it comes in percentages, not integer values (however we could consider them as integers, since they always have exactly two decimal points)
2. we have a separate entry for the total (which is meaningful), and it'd be nice for it to appear at the beginning or the end of the table.

I wonder whether it's best to store the percentages as strings in the artifact. Not sure.


.github/workflows/ci.yml
```diff
@@ -42,6 +42,14 @@ jobs:
           name: stats
           path: entrypoint-sizes.json
           if-no-files-found: error
+      - name: Export the test coverage report
+        run: nix-shell --run 'make ocaml-src && dune runtest --instrument-with bisect_ppx --force . && bisect-ppx-report summary --per-file | python scripts/coverage-to-json.py | tee test-coverage.json'
```
utdemir, line 46: It might be nicer to add this to the `Makefile`, eg. `make test-coverage.json` could depend on `ocaml-src` and run the commands we want (probably then the existing `test-coverage` task would just an alias for `test-coverage.json`).

scripts/coverage-to-json.py
```diff
@@ -0,0 +1,42 @@
...
+                filename = "TOTAL"
+            else:
+                sys.exit("malformed filename in coverage report: {}".format(filename))
+        else:
+            sys.exit("malformed line in coverage report: {}".format(line))
+
+        filedict[filename] = percentage
+
+    json_object = json.dumps(filedict, sort_keys=True, indent=4)
+    print(json_object)
```
utdemir, line 42: There's a trade-off of conciseness here, but we could inline this script to `Makefile` with something like:

```
bisect-ppx-report summary --per-file \
  | sed -E 's/^ *([0-9]+\.[0-9]+) *% *[^ ]* *(.*)$/\1|\2/g' \
  | jq -R 'split("|") | { "value": .[0] | tonumber, "key": (.[1] | if . == "Project coverage" then "TOTAL" else ltrimstr("src/") | rtrimstr(".ml") end) }' \
  | jq -s 'from_entries'
```

Arguably, once one knows a bit about the `jq` syntax this version might even be more readable.

utdemir (commented):
LGTM, and it works! Just a few comments, mainly to reduce repetitions.

> we have a separate entry for the total (which is meaningful), and it'd be nice for it to appear at the beginning or the end of the table.

Yeah, this makes sense. I would be happy with special casing `TOTAL` on `print_diffs`, or making in accept an optional parameter like `total_key=None` which enables that special casing.

.github/workflows/ci.yml
```diff
@@ -42,6 +42,14 @@ jobs:
           name: stats
           path: entrypoint-sizes.json
           if-no-files-found: error
+      - name: Export the test coverage report
```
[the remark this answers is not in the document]
gkaracha: That's a good point. I guess we could run the testsuite only once and have it be instrumented using `bisect_ppx`, but I am a little scared to do that, because then we'd rely on the correctness of `bisect_ppx` more than I'd like—it does alter the code after. If you are not worried about this at all, we could indeed make `nix-build -A michelson --arg doCheck true` always instrument with `bisect_ppx`.

.github/workflows/ci.yml
gkaracha: Ah yes, this command ended up being way too long. Currently our `test-coverage` target also generates the output as html, but I could have two targets, and share the common tasks. Then the one would finish with generating the html output and other would finish by generating json. I'll fix this :+1:

scripts/coverage-to-json.py
gkaracha: Haha, live and learn! Yeah, `jq` seems good for this kind of task, though the conditional in this case makes it a little less appealing to (the future) me (if I were less thorough with the error messages and the comments I think I could probably make the python code fit in 5 lines too :stuck_out_tongue_closed_eyes:). I'll use your `jq` script, since it saves us from having another file in `scripts/` :+1:

PR #216, opened 2021-07-28
Rename Burrow.last_touched to Burrow.last_checker_timestamp

gkaracha:
Since it's original name and description were incorrect.

### DIFF
--- a/src/burrow.ml
+++ b/src/burrow.ml
@@ -30,8 +30,8 @@ type burrow =
      * purposes, this collateral can be considered gone, but depending on the
      * outcome of the auctions we expect some kit in return. *)
     collateral_at_auction : Ligo.tez;
-    (* The last time the burrow was touched. *)
-    last_touched : Ligo.timestamp;
+    (* The timestamp checker had the last time the burrow was touched. *)
+    last_checker_timestamp : Ligo.timestamp;
   }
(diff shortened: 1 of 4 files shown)


utdemir (approved):
LGTM! 

Not that it matters, but am I correct to assume that the gas cost difference is only caused by `last_checker_timestamp` string being longer than `last_touched`, because they appear as annotations on generated code? Or is it just a result of non-deterministic compilation/execution?

gkaracha:
Hmmm, I don't know, actually! When I fired that CI yesterday I didn't expect to see any differences so I am a little surprised about this too. Especially since the entrypoint sizes did not change. I restarted the CI, to see if we can get different results, but I'll look into gas calculation too :+1:

gkaracha:
> Not that it matters, but am I correct to assume that the gas cost difference is only caused by last_checker_timestamp string being longer than last_touched, because they appear as annotations on generated code? Or is it just a result of non-deterministic compilation/execution?

The second run seems to give consistent results with the first run, so I'm inclined to think that it's because of the bigger annotations :thinking: I'll merge this one, but I am now wondering about other field names such as `min_kit_for_unwarranted` and perhaps other names relating to the liquidation auctions. Nah, probably the difference is negligible :slightly_smiling_face:

PR #217, opened 2021-07-30
Replace "nix-shell" by "nix-shell --pure" on CI

gkaracha:
So that we can reproduce the results observed on CI locally.


gkaracha:
@utdemir you were totally right, the `accessToken` was the issue, passing `--keep GITHUB_TOKEN` to the nix shell worked like a charm :ok_hand: In retrospect, I don't think we didn't need `--pure` on the action that only runs within a GitHub context, but uniformity is always nice to have; now we only invoke `nix-shell` with `--pure` in all places :slightly_smiling_face:

PR #218, opened 2021-07-30
Use zero as an absorbing bound for the approximation of the total kit outstanding

gkaracha:
Since I've been unable to prove that our approximation of outstanding kit (i.e., `parameters.outstanding_kit`) overapproximates the real value (i.e., the sum of `burrow.outstanding_kit` for all burrows `burrow`), it is theoretically possible that checker can get stuck, along the lines of #209.

To avoid that, this PR makes sure that we never force `parameters.outstanding_kit` below zero.
- Disadvantage: this catch-all behavior could potentially hide other bugs.
- Advantage: if indeed this case is triggered, keeping `parameters.outstanding_kit` at zero while `burrow.outstanding_kit` gets decreased actually decreases the discrepancy between the real value and the approximation.

Also it adds a unit test to check the new behavior (I don't know how / if it is possible to trigger this execution path using a real-life scenario).

### DIFF
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -221,14 +221,7 @@ let entrypoint_burn_kit (state, (burrow_no, kit): checker * (Ligo.nat * kit)) : 
    * that is not owned by the sender). Either way I (George) think that this
    * assertion is only useful for our unit tests. *)
 
-  let outstanding_to_remove =
-    (* BEGIN_OCAML *)
-    let _ =
-      if gt_kit_kit actual_burned state.parameters.outstanding_kit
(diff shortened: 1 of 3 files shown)


utdemir (approved):
LGTM!

gkaracha:
@utdemir sorry for the late change, but I thought it's best to know if the changes in this PR end up making any difference (I'll probably keep looking into proving the changes necessary/unnecessary to be honest), so I added a couple of OCaml-only warnings if scaling down happens (see commit b8b7d621fa8a2166dd60234c71be10fb5c9f8b29). Feel free to revert if you don't find it necessary. I am otherwise done with this one so feel free to merge at will :slightly_smiling_face:

Issue #153, opened 2021-06-08
Missing tests for liquidation cancellation


gkaracha:
Due to the specific conditions needed for the cancellation of a liquidation, we currently have no test where `Checker.entrypoint_cancel_liquidation_slice` succeeds in returning a liquidation slice back to its burrow. Now that it's been getting easier to write e2e tests it's probably time to add tests for more intricate scenaria where cancellation is possible.

dorranh:
@gkaracha, would you consider this issue resolved after #193, or would you prefer that we add additional tests for this case?

gkaracha:
@dorranh As far as the e2e tests are concerned I do consider it resolved indeed! I would still like us to be able to expand our tests on the OCaml side though (see discussion with @utdemir here: https://github.com/tezos-checker/checker/pull/195#pullrequestreview-705694421). We could resolve this issue and open another one, though I think I'd just leave this one open instead.

PR #219, opened 2021-07-30
Additional unit tests for operations returned by entrypoints

dorranh:
Similar to #214, this PR adds unit tests for operations emitted by entrypoints in `checker.ml`. Some of these tests also overlap with assertions in the `can complete a liquidation auction` test, but I think it is cleaner to have individual unit tests for them as well.

To be able to efficiently write these, I ended up adding a group of test fixture-like functions in testChecker.ml for producing instances of checker with different liquidation auction states. I think these will prove more generally useful as we add more unit tests related to liquidation auctions.

Entrypoints covered include:
- entrypoint_set_burrow_delegate
- entrypoint_receive_price
- entrypoint_mark_for_liquidation
- entrypoint_cancel_liquidation_slice
- entrypoint_liquidation_auction_place_bid
- entrypoint_touch_liquidation_slices
- entrypoint_liquidation_auction_claim_win

### DIFF
--- a/tests/testChecker.ml
+++ b/tests/testChecker.ml
@@ -479,11 +479,11 @@ let suite =
 
        (* Place a bid *)
        Ligo.Tezos.new_transaction ~seconds_passed:10 ~blocks_passed:1 ~sender:alice_addr ~amount:(Ligo.tez_from_literal "0mutez");
-       let _ops, _checker = Checker.entrypoint_liquidation_auction_place_bid
+       let ops, _checker = Checker.entrypoint_liquidation_auction_place_bid
            (checker,
             ((Option.get checker.liquidation_auctions.current_auction).contents, auction_details.minimum_bid))
        in
-       ()
+       assert_operation_list_equal ~expected:[] ~real:ops
     );
 
     ("entrypoint_mark_for_liquidation - emits expected operations" >::
--- a/tests/testChecker.ml
+++ b/tests/testChecker.ml
@@ -43,8 +43,9 @@ let empty_checker_with_cfmm (cfmm: CfmmTypes.cfmm) =
   Checker.assert_checker_invariants checker;
   checker
 
-(* TODO: Test fixtures for checker with different auction states *)
-
+(* Produces a checker state with liquidatable burrows.
+ * Returns a list of the liquidatable burrow ids, underburrowed burrow ids, and the contract state
+*)
 let checker_with_liquidatable_burrows () =
   let checker = empty_checker in
   (* Create some burrows and mint some kit *)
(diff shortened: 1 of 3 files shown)


tests/testChecker.ml
```diff
@@ -325,6 +464,93 @@ let suite =
...
+       Ligo.Tezos.new_transaction ~seconds_passed:0 ~blocks_passed:0 ~sender:alice_addr ~amount:(Ligo.tez_from_literal "0mutez");
+       let _, checker = Checker.entrypoint_mint_kit (checker, (new_burrow_no, auction_details.minimum_bid)) in
+
+       (* Place a bid *)
+       Ligo.Tezos.new_transaction ~seconds_passed:10 ~blocks_passed:1 ~sender:alice_addr ~amount:(Ligo.tez_from_literal "0mutez");
+       let _ops, _checker = Checker.entrypoint_liquidation_auction_place_bid
+           (checker,
+            ((Option.get checker.liquidation_auctions.current_auction).contents, auction_details.minimum_bid))
+       in
+       ()
```
gkaracha, line 486: Instead of ignoring the `ops` here we could check that they are indeed empty, like so:
```suggestion
       let ops, _checker = Checker.entrypoint_liquidation_auction_place_bid
           (checker,
            ((Option.get checker.liquidation_auctions.current_auction).contents, auction_details.minimum_bid))
       in
       assert_operation_list_equal ~expected:[] ~real:ops
```

gkaracha: Also, since we've come this far (this applies to the tests that follow as well), we might as well check that the whole effect of the operation called is as expected (i.e., inspect the resulting `checker` too, and for example in this case check that the current winning bid is the one we placed).

tests/testLib.ml
```diff
@@ -134,3 +136,15 @@ let debug_print_all_kit_in_sealed_state msg wrapper =
...
+let avl_leaves_to_list (mem: Mem.mem) (AVLPtr ptr) : leaf_ptr list =
+  let rec go ptr: leaf_ptr list =
+    match Mem.mem_get mem ptr with
+    | Root (None, _) -> []
+    | Root (Some ptr, _) -> go ptr
+    | Leaf _ ->
+      [LeafPtr ptr]
+    | Branch branch ->
+      List.append (go branch.left) (go branch.right) in
+  go ptr
```
gkaracha, line 150: Ah, yes, another candidate for #150 perhaps.

gkaracha (commented):
Nice! I had a quick look and everything looks in order. I think we might be able to do a little better though:
* Since we are already checking that the operations returned are as expected, we might as well check that the resulting state is as expected. Setting up the environment is quite some work and it'd be nice to reuse that.
* I wonder if we should clarify a little more the properties of the generated state from the test fixtures (e.g., how much tez do we expect each burrow to have, who are the owners, etc.). Not entirely sure how much specificity will be useful. I'm also OK keeping it as-is, I am just wondering how easy it will be to reuse these fixtures in the future :thinking: 

Also seeing how you access slice information from the state reminded me of #150; probably we can/must separate these operations into reusable offline views, though this certainly is work for another PR.

tests/testChecker.ml
gkaracha: Actually, now that I think about it, this pattern of testing only parts of the output shows up in several places in our testsuite. We can consider changing this, but it's not part of this PR. I'm resolving this one.

gkaracha (approved):
LGTM, I'll go ahead and merge this one.

gkaracha:
> Also, since we've come this far (this applies to the tests that follow as well), we might as well check that the whole effect of the operation called is as expected (i.e., inspect the resulting `checker` too, and for example in this case check that the current winning bid is the one we placed).

Actually, now that I went through our issue tracker I saw that we have #153 about this. If we enhance the tests here @dorranh, we can probably finally close that issue.

dorranh:
Awesome, thanks for the review and assist on this while I was out!

PR #223, opened 2021-08-05
Add git to shell

utdemir:
Because `mutate.py` depends on it.

I saw it on the actions log: https://github.com/tezos-checker/checker/actions/runs/1099652310

It's a bit of a shame that we don't have tests for it, but at least it is visible.


gkaracha:
Ah, it must be #217 that broke it, I totally missed it :confused:

gkaracha (approved):
LGTM! I tried it locally and it seems to work as expected :ok_hand:

PR #226, opened 2021-08-09
Add a comment workflow

utdemir:
This PR changes how our PR-commenter-bot-thing works.

Previously, it was a task running on `pull-request` events. This was causing multiple problems:

* Even though the workflow was triggered by `pull-request` events, it was looking at the artifacts of the `build` events, which was a bit unusual.
* Workflows triggered by `pull-request` events could finish before `build` events.
* There might even be cases where a previous commits finishes after a later one, which complicates things a bit.

Overall, I did not have a great experience with things happening tied to the events emitted. So I decided to follow a more declarative approach; where we have a task that:

1. Looks at open PR's
2. Checks if the tests have passed (both of the PR head and base commit)
3. Checks if it already posted a comment about that commit.
4. Otherwise calculates the diff, and sends a comment.

It'd be easy to modify the script to update/delete existing comments.

And here are some disadvantages:

1. It always runs the code on master.
2. It is a scheduled task (every 10 minutes), because the account the bot runs at does not have permission to trigger new workflows at the end of `pull-request` or `build` tasks (that was my initial plan). This is to avoid triggering recursive workflows, GitHub says. But the script does not enter our `nix-shell` unless necessary, so the frequent runs should be fast (it has to install Nix tho, so it'd still take a few minutes)

I tried it until line 68 and had to fix a ton of annoying errors in the process. Then I made it a cronjob instead, and that requires me to get the workflow in `master` in order to try the rest. So it is likely to break a few times in `master`, and then we should try to fix it there.

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -1,0 +1,93 @@
+name: Comment PR results
+
+# only run one instance at a time
+concurrency: bot-mutex
+
+on:
+  schedule:
+    - cron:  '0/10 * * * *'
+  workflow_dispatch: {}
+
+jobs:
+  mutate:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: cachix/install-nix-action@v13
+      - uses: actions/checkout@v2
+        with:
+           ref: 'ud/fix-bot' # TODO: remove before merging
+      - uses: cachix/cachix-action@v10
+        with:
+          name: tezos-checker
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+      - uses: actions/github-script@v4.0.2
+        with:
+          script: |
+            const util = require('util');
+            const execFile = util.promisify(require('child_process').execFile);
+
+            const prs = (await github.pulls.list({
+              owner: context.repo.owner,
+              repo: context.repo.repo,
+              state: "open"
+            })).data;
+            console.log(prs);
+
+            async function testsPassed(ref) {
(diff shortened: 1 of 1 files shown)


.github/workflows/comment.yml
```diff
@@ -0,0 +1,93 @@
...
+              console.log("Processing PR", pr.id);
+
+              const base = pr.base.sha;
+              const head = pr.head.sha;
+
+              // skip if the tests are pending or failed
+              if(!(await (testsPassed(base))) || !(await (testsPassed(head)))) {
+                console.log("Tests not successful, skipping.");
+                return;
+              }
```
gkaracha, line 68: Unless I misunderstood how will the script run, I think that `return` is perhaps too pessimistic here; if the tests for a PR have failed we want to move on the the next one, right? Then `continue` would be more appropriate, since we're in a `for`-loop:

```suggestion
              // skip if the tests are pending or failed
              if(!(await (testsPassed(base))) || !(await (testsPassed(head)))) {
                console.log("Tests not successful, skipping.");
                continue;
              }
```

gkaracha (commented):
Overall it looks good to me, apart from the `return` vs. `continue` comment I left below; do let me know if I got that totally wrong, but otherwise feel free to merge :slightly_smiling_face:

.github/workflows/comment.yml
utdemir: You're absolutely right, this was a mistake on my part, thanks :)!

Issue #213, opened 2021-07-27
Allow FA1.2/FA2 collaterals for minting


utdemir:
From the original issue #171:

> allow FA1.2 / FA2 token as collateral for burrows instead of tez only. A single form of collateral is still accepted for a given checker instance, but different instances can have different collateral

Currently, Checker only allows `tez` as a collateral for minting kit. These `tez` end up in "burrow" contracts that are still controlled by checker, but delegation rights can be set individually.

This issue is to allow different tokens as collaterals instead of `tez`. Likely, `burrow`'s won't be necessary for that case.

utdemir:
One way to implement this would be via conditional compilation. We can abstract away collateral handling to a separate module, and conditionally compile different versions of it based on whether the collateral is `tez` or a token.

Another way I can think of is to have a wrapper FA2 contract for `tez`. We can modify `checker` to only allow a FA2 (or FA1.2) token for collateral, removing all the logic regarding burrows. Then, we can write a `FA2` wrapper for `tez`, and move the burrow logic inside that contract to allow delegation. Then, if I am not missing anything, that wrapper used with checker-with-fa2-collateral contract will have feature parity with the current checker.

The advantage would mainly be avoiding conditional compilation, and shrinking the main contract complexity a bit (since those two components can be tested separately).

utdemir:
Using a FA2 token as a collateral would also require us to obtain the "collateral token / tez" price, so we can use the existing CFMM (which is ctez-kit) and the oracle feed. As far as I know, there is no standard way obtain this information.

Unless we use the collateral token instead of `tez` everywhere, including the CFMM and the oracle feed.

@murbard, what do you think? Is there a convention for getting a FA2 tokens price in tez, or should we also make the CFMM work with the collateral token?

gkaracha:
Another thing I am not entirely clear about: What about the creation deposit? Do we want it to have the same type as the collateral or is it always going to be tez?

gkaracha:
Pasting here some more thoughts on this. Specifically the differences that I can currently see between having tez and having FA2 as collateral.

Unaffected
---
These entrypoints do not interact with collateral in any way so they are
unaffected by the change of the type of collateral:

* `buy_kit`
* `sell_kit`
* `add_liquidity`
* `remove_liquidity`
* `receive_price`
* `transfer`
* `balance_of`
* `update_operators`
* `touch_burrow`
* `mint_kit`
* `burn_kit`
* `liquidation_auction_place_bid`

[...]

PR #228, opened 2021-08-09
Make the core logic deal with tok instead of tez

gkaracha:
Though this transition is pretty menial, it is a necessary first step for addressing #213, independently of the path we end up choosing (conditional compilation vs. having a tez/FA2 wrapper contract). By making all the core logic deal with an opaque-ish type `tok` we can defer the choice of collateral type to a higher level. With this PR `checker.ml` deals with tez—as it did before—and makes calls to `tok_of_tez` before calling core logic and `tez_of_tok` when interpreting results from core logic. No other module makes calls to these conversion functions (excluding the tests ofc).

While working on this I also noticed that our API uses `tez` in names which we probably want to rename to something more generic: `Deposit_tez` and `Withdraw_tez` (which I think we could rename to `Deposit_collateral` and `Withdraw_collateral`).


utdemir (approved):
LGTM :). I fixed a trivial issue and rebased, so if you want to make changes you might need a hard reset.

gkaracha:
> LGTM :). I fixed a trivial issue and rebased, so if you want to make changes you might need a hard reset.

Thanks for having a look! :pray: No more changes I think, this PR is large enough as it is; I'll go ahead and merge it :slightly_smiling_face:

Issue #167, opened 2021-06-24
Missing descriptions in the spec for liquidation-auction-related entrypoints


gkaracha:
We should extend the ["Liquidation Auctions"](https://checker.readthedocs.io/en/latest/functional_spec.html#liquidation-auctions) section in the spec to provide for each entrypoint (a) an one-line description of what the entrypoint does and (b) a brief description of the entrypoint's parameters.

gkaracha:
Closed by #231.

PR #231, opened 2021-08-11
Revamp docs/spec/functional_spec.rst

gkaracha:
The spec was somewhat out of date; this PR addresses this.
* Missing descriptions added for
  - [X] `touch`
  - [X] `receive_price`
  - [X] `receive_slice_from_burrow`
  - [X] `liquidation_auction_claim_win`
  - [X] `liquidation_auction_place_bid`
  - [X] `cancel_liquidation_slice`
  - [X] `touch_liquidation_slices`
  - [X] `mark_for_liquidation`
  - [X] `touch_burrow`
* Type signatures updated for
  - [X] `deactivate_burrow`
  - [X] `burrow_max_mintable_kit`
  - [X] `is_burrow_liquidatable`
  - [X] `is_burrow_overburrowed`
* Notes added on entrypoints that receive tokens, to clarify that they will fail if the transfers needed are authorized and valid.
* References to `muctez` are removed; we use whatever denomination ctez specifies and we've already removed references to mulqt and mukit anyway.
* The type of `ptr` has been `nat` (instead of `int`) for some time now but we'd forgotten to update the spec. This is now fixed.
* Misc consistency fixes.

Also I noticed that we don't explicitly say in the spec that for several entrypoints we fail if `Tezos.amount <> 0`. On one hand I think this is a bonus so we don't have to be so explicit about it, but on the other we do it pretty consistently and perhaps we'd like it to be a part of the spec? Not sure, but we should probably discuss this.

Also addresses #167.


utdemir:
re. the above comment.

On our build workflows, jt seems like we have:

```
  pull_request:
    paths-ignore:
      - '**/*.md'
      - 'docs/**/*'
      - '.readthedocs.yaml'
  push:
    paths-ignore:
      - '**/*.md'
      - 'docs/**/*'
      - '.readthedocs.yaml'
```

So the commits does not have the build artifacts for documentation changes, and our script interprets it as everything being deleted. We should probably fix the script to ignore commits if artifacts are not there.

docs/spec/functional_spec.rst
```diff
@@ -95,52 +96,63 @@ or if the sender is not the burrow owner.
...
 +---------------+-----------------------+-------------------------------------------------------------------------+
 
 
 Perform burrow maintenance
 --------------------------
 
+Perform maintenance tasks on a burrow (i.e., update it's outstanding kit
+according to the system changes that have taken place since the last time the
+burrow was operated on). Fails if the burrow does not exist.
+
```
utdemir, line 140: Maybe it's worth noting that most burrow operations call this implicitly and this is only here as a failsafe mechanism.

docs/spec/functional_spec.rst
```diff
@@ -244,46 +259,146 @@ Liquidation Auctions
...
+Perform housekeeping tasks on the contract state. This includes:
+
+#. updating the system parameters;
+#. accruing burrowing fees to the cfmm;
+#. updating auction-related info (completing an old / starting a new auction);
+#. processing a limited number of liquidation slices from completed auctions;
+#. updating the index by consulting the oracle.
+
+This operation credits an amount of kit (that is a function of time passed
+since the last time ``touch`` was called) to ``Tezos.sender``.
```
utdemir, line 377: To me, it would be clearer to have the word "reward" somewhere in the description, maybe something like:

```suggestion
since the last time ``touch`` was called) as a reward to ``Tezos.sender``.
```

docs/spec/functional_spec.rst
```diff
@@ -244,46 +259,146 @@ Liquidation Auctions
...
 ``receive_price: nat``
 
++---------------+-----------------------+-------------------------------------------------------------------------+
+| Parameter     |      Field Type       | Description                                                             |
++===============+=======================+=========================================================================+
+| price         | nat                   | The current index, as a fixedpoint with a scaling factor of 1000000     |
++---------------+-----------------------+-------------------------------------------------------------------------+
+
+
 FA1.2 Interface
```
utdemir, line 402: Not about this PR, but this should probably be called "FA2 Interface".

```suggestion
FA2 Interface
```

utdemir (approved):
LGTM. I just had a read through and noted a few suggestions, but none of them is really about this PR, so we might do it as part of another PR if you prefer that.

docs/spec/functional_spec.rst
gkaracha: Ah, yes, I'll fix that, good catch :+1:

gkaracha:
@utdemir re. the comments on touching and the reward: the way I understood it the "functional" spec is to be more about _what_ happens, rather than _why_ it happens, that's why I didn't add more text about the rationale. Personally I'd also like some more elaboration on the system but I am not sure if this is the place for it. On the other hand, we don't have the calculations in this document either (which would make it more "functional") so I am not sure what should and what should not be included in this file :thinking: @purcell any thoughts?

gkaracha:
> re. the above comment.
> 
> It seems like we are skipping our `build` workflows if only changed files are documentation:
> 
> ```
>   pull_request:
>     paths-ignore:
>       - '**/*.md'
>       - 'docs/**/*'
>       - '.readthedocs.yaml'
>   push:
>     paths-ignore:
>       - '**/*.md'
>       - 'docs/**/*'
>       - '.readthedocs.yaml'
> ```
> 
> So the commits does not have the build artifacts for documentation changes, and our script interprets it as everything being deleted. We should probably fix the script to ignore commits if artifacts are not there.

Now that I think about it, I think I'd prefer if we always had artifacts for `master`. Might be a little wasteful on some occasions, but I think it's safer and more consistent; I'd remove the following part from `.github/workflows/ci.yml`:
```
  push:
    paths-ignore:
      - '**/*.md'
      - 'docs/**/*'
      - '.readthedocs.yaml'
```

purcell:
> Personally I'd also like some more elaboration on the system but I am not sure if this is the place for it. On the other hand, we don't have the calculations in this document either (which would make it more "functional") so I am not sure what should and what should not be included in this file 🤔 @purcell any thoughts?

Things that are internal details probably don't belong in the document, as far as I am concerned. The endpoint documentation so far serves primarily as a user-level / API-level guide, ie. the information needed  to use Checker. I'm generally wary of writing too much about the code in the documentation if it's not helpful to a specific target reader, because the two will inevitably get out of sync.

PR #233, opened 2021-08-11
ci: Catch errors on the comment workflow

utdemir:
`artifacts.py` might fail if the artifacts provided by the job is not of the correct shape. If this happens, it wouldn't be nice for the global comment task to fail. So, this PR wraps it with a `try..catch`. Unfortunately, as usual, we can not test it on this branch, so we'll see if it fails after merging.

### DIFF
--- a/.github/workflows/comment.yml
+++ b/.github/workflows/comment.yml
@@ -74,20 +74,29 @@ jobs:
               // only run the job if there is no existing comment
               if(await (commentExists(pr.number, marker))) {
                 console.log("Comment exists, skipping.");
-              } else {
-                const ret = await (execFile(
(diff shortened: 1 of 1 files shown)


gkaracha (approved):
It looks good to me, at least as far as I can see. When you're ready, feel free to land it on master and we can debug there :sweat_smile:

By the way, has this happened already?
> `artifacts.py` might fail if the artifacts provided by the job is not of the correct shape. If this happens, it wouldn't be nice for the global comment task to fail.

utdemir:
> By the way, has this happened already?

As far as I know, no. I just had a look at scripts failure cases and noticed this.

PR #234, opened 2021-08-12
Add a docker-based workflow

utdemir:
This PR introduces a Docker-based workflow for developing checker.

It provides instructions to build a Docker image which contains Nix alongside with a prepopulated `/nix/store` with Checker's dependencies. It expects a Checker repository to be mounted under `/nix`, and automatically enters to the `nix-shell` on entrypoint.

We had a chat with @gkaracha and decided to remove VSCode from the mix and just provide a simple Docker image. They can still edit the code using vscode, and see the errors by running `make` under the docker container (Or they can do `fd | entr -c make build-ocaml` which is exactly the same way that I develop checker, so in my opinion it is not horrible :) ).

I considered a GitHub action to build and test this Docker image, but then I decided against it because I don't think this will be changed frequently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,21 @@ For running the end-to-end tests, type:
 $ CHECKER_DIR=$(nix-build -A michelson --arg e2eTestsHack true --no-out-link) python e2e/main.py
 ```
 
+### Development using Docker
+
+If you are not on Linux, or do not want to install the Nix package manager; a Docker-based development environment is also
+provided:
+
+```
+$ docker build --build-arg USER_UID=$(id -u) --build-arg USER_GID=$(id -g) -f nix/Dockerfile . -t checker-devcontainer
+$ docker run -it -v "$PWD:/mnt" checker-devcontainer
+```
+
+Within the image, you can use the `make` based workflow above. It also contains the `checker` executable that lets you
+deploy the contract.
+
+If you change the dependencies (including the code for the `checker` executable), spin up a fresh container.
+
 ## Local Deployment
 
 The contract can be deployed to a local, Docker sandbox run using the provided
--- a/README.md
+++ b/README.md
@@ -65,7 +65,7 @@ $ CHECKER_DIR=$(nix-build -A michelson --arg e2eTestsHack true --no-out-link) py
 If you are not on Linux, or do not want to install the Nix package manager; a Docker-based development environment is also
 provided:
 
-```
+```console
 $ docker build --build-arg USER_UID=$(id -u) --build-arg USER_GID=$(id -g) -f nix/Dockerfile . -t checker-devcontainer
 $ docker run -it -v "$PWD:/mnt" checker-devcontainer
 ```
--- a/nix/Dockerfile
+++ b/nix/Dockerfile
@@ -1,0 +1,62 @@
+FROM ubuntu:21.10
+
+RUN apt-get update
+RUN apt-get install -y \
+  curl xz-utils sudo \
+  # brings /etc/services which seems to be necessary for Nix
+  netbase
+
+# https://aka.ms/vscode-remote/containers/non-root-user
+ARG USER=checkeruser
+ENV USER=$USER
+
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+RUN groupadd -f --gid $USER_GID $USER \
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -60,6 +60,21 @@ For running the end-to-end tests, type:
 $ CHECKER_DIR=$(nix-build -A michelson --arg e2eTestsHack true --no-out-link) python e2e/main.py
 ```
 
+### Development using Docker
+
+If you are not on Linux, or do not want to install the Nix package manager; a Docker-based development environment is also
+provided:
+
+```
```
gkaracha, line 68:
```suggestion
```console
```
Just to look a little prettier :smile:

gkaracha (approved):
LGTM! I followed the instructions and it all works OK on my machine :ok_hand:

One concern I have (but that's more generally a docker issue I guess) is that I couldn't edit the files inside the image since no editor was available.

Let's wait to see what @purcell says before merging, since he's not on Linux.

utdemir:
> One concern I have (but that's more generally a docker issue I guess) is that I couldn't edit the files inside the image since no editor was available.

My intention is that you'd edit the files outside of the container, and since the host directory is mounted inside the container the changes will be seen inside the container too. Maybe we should clarify it on the README.

gkaracha:
> My intention is that you'd edit the files outside of the container, and since the host directory is mounted inside the container the changes will be seen inside the container too. Maybe we should clarify it on the README.

Ah, right, I didn't think of that :facepalm: Yes, I can confirm that this approach works. Perhaps add that line to the README in case others are as absent-minded as I am :sweat_smile:

gkaracha:
@utdemir I thought I'd try out the end-to-end tests too
```console
$ CHECKER_DIR=$(nix-build -A michelson --arg e2eTestsHack true --no-out-link) python e2e/main.py
```
but it seems to fail. You can see the whole output here: https://gist.github.com/gkaracha/fc49146d3ead446240c6907cdf26a354

utdemir:
Oh, right. That likely is because the Docker socket the clients use to connect to the system docker daemon is not available inside the container. Could you try with passing

```
-v /var/run/docker.sock:/var/run/docker.sock
```

to `docker run`?

gkaracha:
> Oh, right. That likely is because the Docker socket the clients use to connect to the system docker daemon is not available inside the container. Could you try with passing
> 
> ```
> -v /var/run/docker.sock:/var/run/docker.sock
> ```
> 
> to `docker run`?

You mean like this?
```console
$ docker run -it -v /var/run/docker.sock:/var/run/docker.sock -v "$PWD:/mnt" checker-devcontainer
```
Gives me the same error :thinking:

utdemir:
> Gives me the same error

It isn't quite the same error :). The previous one was:

```
urllib3.exceptions.ProtocolError: ('Connection aborted.', FileNotFoundError(2, 'No such file or directory'))
```

But the new one is:

```
urllib3.exceptions.ProtocolError: ('Connection aborted.', PermissionError(13, 'Permission denied'))
```

The previous error was about that socket not existing on the docker container. Turns out the new one is a permission error, where `/var/run/docker.sock` is only has `rw` permissions for either the `root` user or the `docker` group. SInce we're not `root`, and the `docker` group does not exist on the container, we can not use the socket. I guess one solution would be to add a `docker` group to the container and ensure that it has the same `gid` of the `docker` group on the host machine.

However, I do think we can do it later. We're only missing e2e tests, and they are not very useful for development since they takes ages. One can always push a commit and let the CI do it when absolutely necessary.

gkaracha:
> It isn't quite the same error :).

Ah, sorry, you're absolutely right. I skimmed the output too fast :facepalm: 

> However, I do think we can do it later. We're only missing e2e tests, and they are not very useful for development since they takes ages. One can always push a commit and let the CI do it when absolutely necessary.

Yeah, doing it later sounds good to me, it's certainly not a priority. I would change the README a little though, to be clear that we are aware of this limitation. Something like this perhaps:
```diff
- Within the image, you can use the `make` based workflow above.
+ Within the image, you can use the `make` based workflow above (but not run the e2e tests, currently).
```

dorranh:
@utdemir, I'm going to go ahead and close this since I think it was superseded by #245. Please ping me and re-open the PR if there are changes which aren't covered by #245 though. :slightly_smiling_face:

PR #235, opened 2021-08-13
Add a GitHub workflow to deploy checker to a testnet

utdemir:
This PR adds a GitHub Actions workflow to deploy checker to a testnet.

Once this is merged, someone with write access to the checker repository should be able to run this workflow (via navigating to this workflow through the "Actions" tab), and it should even ask for some inputs like the checker branch to deploy, RPC node, and ctez and oracle contracts.

I also withdrew some funds from the faucet to a fresh account, and created a secret called `TESTNET_ACCOUNT_SECRET_KEY` that is used by this workflow. It is not supposed to be accessible if one does not have write access to this repository, and even if it leaks somehow it is not the end of the world.

You can see (in half an hour or so) an example run here: https://github.com/tezos-checker/checker/actions/runs/1127054476

### DIFF
--- a/.github/workflows/deploy-testnet.yml
+++ b/.github/workflows/deploy-testnet.yml
@@ -16,15 +16,15 @@ on:
       node:
         description: 'Node to connect to'
         required: true
-        default: 'https://edonet.smartpy.io/'
+        default: 'https://rpczero.tzbeta.net/'
       ctez:
         description: 'Ctez contract'
         required: true
-        default: 'KT1JHiNYzzwYZ2sZmZGcPCqgi3RAjoWshntQ'
+        default: 'KT1XaQyVJHKzr1qXPMoLawi56esqKzyzeuhE'
       oracle:
         description: 'Oracle contract'
         required: true
-        default: 'KT1RAcdY2zaX5PryXV9PDdJsT3cubWXkVEBg'
+        default: 'KT1Xbo2Lkg6nPqyka4fqD7kLTtW5wDD3vkGe'
 
 jobs:
   deploy:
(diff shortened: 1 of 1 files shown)


.github/workflows/deploy-testnet.yml
```diff
@@ -0,0 +1,61 @@
+# This workflow depends on a repository secret called `TESTNET_ACCOUNT_SECRET_KEY`.
+# It contains the private key on a testnet. To reproduce, fetch an account from the faucet,
+# and run:
+#
+#   tezos-client activate account alice with tz1xxx.json
+#   tezos-client reveal key for alice
+#   tezos-client show address alice -S
+#
+# Then set the `Secret Key` as the aforamentioned secret. It should be of form: unencrypted:edsk****
```
gkaracha, line 9:
```suggestion
# Then set the `Secret Key` as the aforementioned secret. It should be of the form: unencrypted:edsk****
```

gkaracha (approved):
Neat! :ok_hand: I wonder if it's a good idea to have defaults for `ctez` and `oracle` given how easily they can get out of sync (as we saw yesterday). Not important really, I'd even merge this as-is. :+1:

.github/workflows/deploy-testnet.yml
utdemir: TIL, thanks :)

utdemir:
> I wonder if it's a good idea to have defaults for ctez and oracle given how easily they can get out of sync 

You're right. Protocol updates happen every month or so, so you're right that it's not so nice to have something becoming obsolete that often. However we might also deploy checker more often than that. 

I think for now we can leave it there, and we'll see if updating it every once in a while becomes annoying :).

PR #237, opened 2021-08-14
Update design.rst

Cryptinterested:
I changed block time to 30 seconds since Granada is now live

### DIFF
--- a/docs/spec/design.rst
+++ b/docs/spec/design.rst
@@ -28,7 +28,7 @@ Any computation on a blockchain happens in discrete time. We note the
 timestamps at which the Checker system is updated as a series of
 increasing timestamps, :math:`t_i`. Ideally, these updates happen every
 time a new block is added to the blockchain and thus, under the current
-economic protocol, those timestamps are separated by about a minute
+economic protocol, those timestamps are separated by about 30 seconds
 each.
 
 That said, the system is designed to be resilient to changes in the


utdemir:
Thanks.

This makes me wonder if any of our default constants should also change, specifically: [`max_bid_interval_in_blocks`](https://github.com/tezos-checker/checker/blob/ce612f2d88f2a641ec020fec724695180645019d/src/constants.ml#L93). 

Previously that `max_bid_interval_in_blocks` and `max_bid_interval_in_seconds` were aligned (eg. they would take approximately the same time). To keep the same ratio, it might make sense to double the `max_bid_interval_in_blocks`. What do you think @gkaracha?

gkaracha:
> This makes me wonder if any of our default constants should also change, specifically: [`max_bid_interval_in_blocks`](https://github.com/tezos-checker/checker/blob/ce612f2d88f2a641ec020fec724695180645019d/src/constants.ml#L93).
> 
> Previously that `max_bid_interval_in_blocks` and `max_bid_interval_in_seconds` were aligned (eg. they would take approximately the same time). To keep the same ratio, it might make sense to double the `max_bid_interval_in_blocks`. What do you think @gkaracha?

Ah, yes, great point @utdemir :ok_hand: I certainly agree on keeping the ratio, but I wonder if to achieve that we should half `max_bid_interval_in_seconds` or double `max_bid_interval_in_blocks`. Either option would make sense to me, but the first is more fast-paced than the second (time-wise). This also makes me think about the descending auction price which is also a function of time. If everything happens faster regarding the bids, we might want to make the minimum bid descend a little faster as well, otherwise the first bid might take many blocks to appear.

Issue #241, opened 2021-08-16
Use dockerised versions of Ligo


purcell:
Building Ligo from scratch with Nix is impractical given the upstream abandonment of nix support, so we need to use pre-built Ligo executables. Currently only x86_64 linux binaries are available upstream. This is a particular problem for MacOS users with M1 processors (me, Derek), where Docker prefers ARM images, ie. linux-arm64.

Additionally, we would like to patch our Ligo until https://gitlab.com/ligolang/ligo/-/issues/1227 is resolved.

The suggested way forward is to create a minimal fork of the upstream Ligo repo in this github org, with the patches applied, and then build both arm64 and x86_64 docker images into the github repo's container registry. Building can be done manually for now, offline. Automation of all builds would be hard to achieve due to lack of arm64 support in GitHub Actions.

With images built, we can start using the dockerised version of Ligo in our local builds.

-----

### Done so far:

Created a fork at https://github.com/tezos-checker/ligo, with our patch applied.

To obtain permissions to push new images, follow the guide to set up push access: https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry

To build multi-arch images, see the [`docker buildx` info](https://docs.docker.com/desktop/multi-arch/) as a little initial set-up may be necessary.

```sh
touch changelog.txt
docker buildx build --platform linux/amd64,linux/arm64 -t ghcr.io/tezos-checker/ligo:0.22.0-checker --push .
```

### To-do:

Use the docker image in the build-ligo script. This isn't working for me out of the box:
```
ruby ./scripts/compile-ligo.rb
Compiling contract.
File "/Users/steve/Projects/tweag/nomadiclabs/checker/scripts/../generated/ligo/checker.mligo", line 94, characters 34-51:

Invalid entrypoint value.
The entrypoint value must be a string literal.
Traceback (most recent call last):
./scripts/compile-ligo.rb:26:in `<main>': compile-contract failed: (RuntimeError)
```
^ Have you seen that before, @utdemir?

utdemir:
>  Have you seen that before, @utdemir?

I haven't seen that, but had a look at it and saw that it caused by [a recent ligo change](https://gitlab.com/ligolang/ligo/-/commit/476af3c62bf483606b6e2ee50fd9e1b8c45a7823).

[...]

PR #242, opened 2021-08-16
Run ligo via docker (see #241)

purcell:
Draft, not fully working locally on Darwin.


dorranh:
I believe that this has been superseded by #245. Should we close this issue @purcell?

PR #244, opened 2021-08-17
Add off-chain views for the current liquidation auction

dorranh:
This PR adds an off-chain view which returns the following current auction data (some of which are listed in #150):
- auction_id
- total collateral in auction lot
- current winning bid
- minimum bid
- remaining duration in seconds and blocks

I think that this should be the minimal information a prospective user would need to query to be able to understand and interact with a given "current" auction.

### DIFF
--- a/docs/spec/functional_spec.rst
+++ b/docs/spec/functional_spec.rst
@@ -614,3 +614,19 @@ Seal the contract and make it ready for use
 -------------------------------------------
 
 ``sealContract: (pair address address)``
+
+Off-chain views
+===============
+
+Get details on the current liquidation auction
+----------------------------------------------
+
+Returns an error if there is currently no active liquidation auction.
+
+``currentLiquidationAuctionDetails: unit -> view_current_liquidation_auction_details_result``
+
++---------------+-----------------------+-------------------------------------------------------------------------+
+| Parameter     |      Field Type       | Description                                                             |
++===============+=======================+=========================================================================+
+| unit          | unit                  | ()                                                                      |
++---------------+-----------------------+-------------------------------------------------------------------------+
--- a/docs/spec/functional_spec.rst
+++ b/docs/spec/functional_spec.rst
@@ -577,14 +577,12 @@ Check whether a burrow can be liquidated
 +---------------+-----------------------+-------------------------------------------------------------------------+
 
 
-Minimum bid for the current liquidation auction (if exists)
------------------------------------------------------------
+Get details on the current liquidation auction
+----------------------------------------------
 
-Returns a pair of the unique identifier for the current auction and the minimum
-amount of kit that one could currently bid in said auction. Fails if there is
-no open auction.
+Fails if there is currently no liquidation auction.
 
-``current_liquidation_auction_minimum_bid : unit -> pair nat nat``
+``current_liquidation_auction_details: unit -> view_current_liquidation_auction_details_result``
 
 +---------------+-----------------------+-------------------------------------------------------------------------+
 | Parameter     |      Field Type       | Description                                                             |
@@ -614,19 +612,3 @@ Seal the contract and make it ready for use
 -------------------------------------------
 
 ``sealContract: (pair address address)``
-
-Off-chain views
-===============
-
-Get details on the current liquidation auction
-----------------------------------------------
-
-Returns an error if there is currently no active liquidation auction.
-
-``currentLiquidationAuctionDetails: unit -> view_current_liquidation_auction_details_result``
-
-+---------------+-----------------------+-------------------------------------------------------------------------+
-| Parameter     |      Field Type       | Description                                                             |
-+===============+=======================+=========================================================================+
-| unit          | unit                  | ()                                                                      |
-+---------------+-----------------------+-------------------------------------------------------------------------+
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -958,18 +958,11 @@ let view_is_burrow_liquidatable (burrow_id, state: burrow_id * checker) : bool =
   let burrow = burrow_touch state.parameters burrow in
   burrow_is_liquidatable state.parameters burrow
 
-let view_current_liquidation_auction_minimum_bid ((), state: unit * checker) : view_current_liquidation_auction_minimum_bid_result =
-  assert_checker_invariants state;
-  let auction = liquidation_auction_get_current_auction state.liquidation_auctions in
-  { auction_id = auction.contents
-  ; minimum_bid = liquidation_auction_current_auction_minimum_bid auction
-  }
-
 let view_current_liquidation_auction_details ((), state: unit * checker) : view_current_liquidation_auction_details_result =
   assert_checker_invariants state;
   let auction = liquidation_auction_get_current_auction state.liquidation_auctions in
   let current_bid, blocks, seconds = match auction.state with
-    | Descending _ -> None, None, None
+    | Descending _ -> (None: (bid option)), (None: (Ligo.int option)), (None: (Ligo.int option))
     | Ascending (current_bid, bid_time_seconds, bid_level) ->
       (
         Some current_bid,
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -973,14 +973,21 @@ let view_current_liquidation_auction_id ((), state: unit * checker) : liquidatio
 
 let view_current_liquidation_auction_winning_bid ((), state: unit * checker) : bid option =
   assert_checker_invariants state;
-  match state.liquidation_auctions.current_auction with
-  | None -> (None : bid option)
-  | Some current_auction ->
-    let bid = match current_auction.state with
-      | Descending _ -> (None : bid option)
-      | Ascending (bid, _, _) -> Some bid
-    in
-    bid
+  let current_auction = liquidation_auction_get_current_auction state.liquidation_auctions in
+  match current_auction.state with
+  | Descending _ -> (None: bid option)
+  | Ascending (bid, _, _) -> Some bid
+
+let view_current_liquidation_auction_remaining_duration ((), state: unit * checker) : view_current_liquidation_auction_remaining_duration_result option =
+  assert_checker_invariants state;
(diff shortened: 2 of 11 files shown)


src/checker.ml
```diff
@@ -965,6 +965,36 @@ let view_current_liquidation_auction_minimum_bid ((), state: unit * checker) : v
   ; minimum_bid = liquidation_auction_current_auction_minimum_bid auction
   }
 
+let view_current_liquidation_auction_id ((), state: unit * checker) : liquidation_auction_id option =
+  assert_checker_invariants state;
+  match state.liquidation_auctions.current_auction with
+  | None -> None
+  | Some current_auction -> Some current_auction.contents
```
gkaracha, line 972: I think this one is part of `view_current_liquidation_auction_minimum_bid`, right?

src/checker.ml
```diff
@@ -965,6 +965,36 @@ let view_current_liquidation_auction_minimum_bid ((), state: unit * checker) : v
...
+
+let view_current_liquidation_auction_winning_bid ((), state: unit * checker) : bid option =
+  assert_checker_invariants state;
+  let current_auction = liquidation_auction_get_current_auction state.liquidation_auctions in
+  match current_auction.state with
+  | Descending _ -> (None: bid option)
+  | Ascending (bid, _, _) -> Some bid
```
gkaracha, line 979: `liquidation_auction_get_current_auction` already fails if there is pending auction so I think you could make all these views fail if there is no ongoing auction; no need to deal with `option` types :wink:

src/checker.ml
```diff
@@ -965,6 +965,36 @@ let view_current_liquidation_auction_minimum_bid ((), state: unit * checker) : v
...
+    Some {
+      seconds=Ligo.sub_timestamp_timestamp (Ligo.add_timestamp_int bid_time_seconds max_bid_interval_in_seconds) !Ligo.Tezos.now;
+      blocks=Ligo.sub_nat_nat (Ligo.add_nat_nat bid_level max_bid_interval_in_blocks) !Ligo.Tezos.level;
+    }
+
+let view_current_liquidation_auction_collateral ((), state: unit * checker) : tok option =
+  assert_checker_invariants state;
+  match state.liquidation_auctions.current_auction with
+  | None -> (None: tok option)
+  | Some auction -> Some (avl_tok state.liquidation_auctions.avl_storage auction.contents)
```
gkaracha, line 996: Same here:
```suggestion
  let current_auction = liquidation_auction_get_current_auction state.liquidation_auctions in
  avl_tok state.liquidation_auctions.avl_storage current_auction.contents
```

tests/testLib.ml
```diff
@@ -23,12 +23,22 @@ type key_hash_option = Ligo.key_hash option [@@deriving show]
...
+let assert_bid_equal ~expected ~real = OUnit2.assert_equal ~printer:LiquidationAuctionPrimitiveTypes.show_bid expected real
+type bid_option = bid option [@@deriving show]
+let assert_bid_option_equal ~expected ~real = OUnit2.assert_equal ~printer:show_bid_option expected real
 let assert_liquidation_result_equal ~expected ~real = OUnit2.assert_equal ~printer:Burrow.show_liquidation_result expected real
 let assert_avl_ptr_equal ~expected ~real = OUnit2.assert_equal ~printer:LiquidationAuctionPrimitiveTypes.show_avl_ptr expected real
 let assert_ctez_equal ~expected ~real = OUnit2.assert_equal ~printer:Ctez.show_ctez expected real
 let assert_parameters_equal ~expected ~real = OUnit2.assert_equal ~printer:Parameters.show_parameters expected real (* FIXME: contains a ratio *)
 let assert_liquidation_slice_contents_equal ~expected ~real = OUnit2.assert_equal ~printer:LiquidationAuctionPrimitiveTypes.show_liquidation_slice_contents expected real
 
+(* Note: This name is a real mouthful but follows the convention we are using in checkerTypes.ml *)
```
gkaracha, line 35: Hahaha, yes, yes it is :sweat_smile:

src/checkerTypes.ml
```diff
@@ -56,4 +56,10 @@ type view_current_liquidation_auction_minimum_bid_result =
   }
 [@@deriving show]
 
+type view_current_liquidation_auction_remaining_duration_result =
+  { blocks: Ligo.int
+  ; seconds: Ligo.int
+  }
+[@@deriving show]
```
gkaracha, line 63: I see `view_current_liquidation_auction_minimum_bid_result` and `view_current_liquidation_auction_remaining_duration_result` and all the duplication in the views and I am wondering if it would be better to just have a big `current_liquidation_auction_data` type and one massive view returning everything about it (something like `view_current_auction_data`) :thinking: @dorranh what do you think? If I remember correctly  @utdemir had suggested something along these lines at some point as well.

dorranh: Hmm that's a good point and could make more sense for these current auction views specifically. One thing we might want to watch out for is combining more efficient views with ones which have to do things like traversing the queue since users might want to do things like poll easy-to-lookup data such as the current bid. This is not an issue for this set of views though since we are only performing lookups and some arithmetic. I'll take a look at merging them.

gkaracha: Yeah, good point :+1: I'd say we can at least group the following "cheap" ones regarding the current auction:

* id
* amount of tez being auctioned off
* currently winning bid
* minimum bid to place
* blocks remaining until closing
* seconds remaining until closing

If we group them together then I think your initial approach of using `option` is better, since we don't want to fail to give any result if for example nobody has bid yet (i.e., there is no "currently winning bid"). So, you could get the current auction via `liquidation_auction_get_current_auction` which could fail if there is no current auction, but after that most of the fields in the returned data structure should probably be `option`s.

For the expensive entrypoints I think we could ask them individually indeed

src/checker.ml
```diff
@@ -958,11 +958,25 @@ let view_is_burrow_liquidatable (burrow_id, state: burrow_id * checker) : bool =
...
   let auction = liquidation_auction_get_current_auction state.liquidation_auctions in
-  { auction_id = auction.contents
-  ; minimum_bid = liquidation_auction_current_auction_minimum_bid auction
+  let current_bid, blocks, seconds = match auction.state with
+    | Descending _ -> (None: (bid option)), (None: (Ligo.int option)), (None: (Ligo.int option))
+    | Ascending (current_bid, bid_time_seconds, bid_level) ->
+      (
+        Some current_bid,
+        Some (Ligo.sub_nat_nat (Ligo.add_nat_nat bid_level max_bid_interval_in_blocks) !Ligo.Tezos.level),
+        Some (Ligo.sub_timestamp_timestamp (Ligo.add_timestamp_int bid_time_seconds max_bid_interval_in_seconds) !Ligo.Tezos.now)
```
gkaracha, line 970: I now wonder what the values of `Tezos.level` and `Tezos.now` are in an offline environment where the view is executed :thinking: Might be a better idea to instead give "safer" values such as
```ocaml
last_block_to_bid_in     = Ligo.add_nat_nat bid_level max_bid_interval_in_blocks
last_timestamp_to_bid_in = Ligo.add_timestamp_int bid_time_seconds max_bid_interval_in_seconds
```
To me it's fine either way, but we should probably test at some point that these two work as expected in a non-OCaml environment. In the long run we'd probably prefer the stable versions.

gkaracha (approved):
LGTM! :+1: I left a comment about the design—sorry I didn't think about it last time I had a look—but it's not urgent. Also, don't forget to update `docs/spec/functional_spec.rst` with the API changes.

src/checker.ml
dorranh: That's a good point. After looking into it, it seems like it depends on how you execute the view. pytezos, for example, uses it's own Michelson interpreter along the contract state at whatever level you specify. If you use the interpreter on a tezos node, I think that `Tezos.level` and `Tezos.now` will match whatever level the node is synced to.

In practice you would need the node to be fully synced to make a bid, so I don't think the current approach is too inaccurate. Your suggestion would definitely leave less room for interpretation since it forces the caller to think about what `Tezos.now` actually is.

gkaracha: Right, I thought so. Then I'd certainly suggest we switch to the clearer versions, either in this PR or in the next.

dorranh: Okay great, I think I'll go ahead and merge this one then look into updating the logic in a follow-up PR. This seems to be a bit of a question of convenience (avoiding extra user logic and calls to the node to check the level / time) vs. ensuring accuracy, and I'm honestly not sure which approach would be most valuable to a prospective liquidation auction user :thinking:

gkaracha: Yeah, I guess it is a little extra logic for the call sites, but I think the extra clarity and unambiguity is certainly worth it. If users are participating in liquidation auctions I expect them to be able to figure out the current level/timestamp themselves.

docs/spec/functional_spec.rst
```diff
@@ -577,14 +577,12 @@ Check whether a burrow can be liquidated
...
+Get details on the current liquidation auction
+----------------------------------------------
 
-Returns a pair of the unique identifier for the current auction and the minimum
-amount of kit that one could currently bid in said auction. Fails if there is
-no open auction.
+Fails if there is currently no liquidation auction.
 
-``current_liquidation_auction_minimum_bid : unit -> pair nat nat``
+``current_liquidation_auction_details: unit -> view_current_liquidation_auction_details_result``
```
gkaracha, line 585: @dorranh In the functional spec we've been using Michelson types (that's what runs on the chain so it's the Michelson code that needs a spec). Since we are not using `[@layout:comb]`, `view_current_liquidation_auction_details_result` is most likely represented in Michelson as
```
(pair nat (pair nat (pair nat (pair (option (pair address nat)) (pair (option int) (option int))))))
```
 so this should probably be:
```
``current_liquidation_auction_details: unit -> (pair nat (pair nat (pair nat (pair (option (pair address nat)) (pair (option int) (option int))))))``
```

dorranh: Ah good catch. I'll submit an update.

PR #249, opened 2021-09-15
Update base Docker image

dorranh:
Updates the base Docker image used in our builds / dev container to 20.04 (downgraded from 21.04). This allows commands to properly run on a wider variety of hosts due to the glibc - host compatibility issues introduced in the glibc version which ships with 21.04. 

Also fixed a small error in the CI workflow which was preventing cache pushes from occurring on master.


gkaracha (approved):
LGTM! :+1: Let's just wait for CI to finish before merging.

PR #250, opened 2021-09-15
Update manual / cron workflows

dorranh:
Fixes a syntax error in the "Mutate" scheduled workflow which was preventing it from running. Also updated the "Deploy to the testnet" workflow to use the new client docker image. Sadly I can't test the latter until we merge this into master :slightly_frowning_face: but the changes should hopefully work (gotta love Github Actions).

### DIFF
--- a/.github/workflows/mutate.yml
+++ b/.github/workflows/mutate.yml
@@ -10,10 +10,10 @@ jobs:
     strategy:
       matrix:
         args:
-          - '--build-arg n_mutations=500 --build-arg modules=src/{burrow,cfmm,checker,checkerMain,common,constants}.ml'
-          - '--build-arg n_mutations=500 --build-arg modules=src/{fa2Interface,lqt,parameters}.ml'
-          - '--build-arg n_mutations=50  --build-arg modules=src/{liquidationAuction,sliceList,mem,ptr}.ml'
-          - '--build-arg n_mutations=50  --build-arg modules=src/avl.ml'
+          - '--build-arg n_mutations=500 --build-arg modules="src/burrow.ml src/cfmm.ml src/checker.ml src/checkerMain.ml src/common.ml src/constants.ml"'
+          - '--build-arg n_mutations=500 --build-arg modules="src/fa2Interface.ml src/lqt.ml src/parameters.ml"'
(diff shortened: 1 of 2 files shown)


.github/workflows/mutate.yml
```diff
@@ -10,10 +10,10 @@ jobs:
...
       matrix:
         args:
-          - '--build-arg n_mutations=500 --build-arg modules=src/{burrow,cfmm,checker,checkerMain,common,constants}.ml'
-          - '--build-arg n_mutations=500 --build-arg modules=src/{fa2Interface,lqt,parameters}.ml'
-          - '--build-arg n_mutations=50  --build-arg modules=src/{liquidationAuction,sliceList,mem,ptr}.ml'
-          - '--build-arg n_mutations=50  --build-arg modules=src/avl.ml'
+          - '--build-arg n_mutations=500 --build-arg modules="src/burrow.ml src/cfmm.ml src/checker.ml src/checkerMain.ml src/common.ml src/constants.ml"'
+          - '--build-arg n_mutations=500 --build-arg modules="src/fa2Interface.ml src/lqt.ml src/parameters.ml"'
+          - '--build-arg n_mutations=50  --build-arg modules="src/liquidationAuction.ml src/sliceList.ml src/mem.ml src/ptr.ml"'
+          - '--build-arg n_mutations=50  --build-arg modules="src/avl.ml"'
```
gkaracha, line 16: By the way I noticed yesterday that if you put the arguments _after the target_ you don't need the `--build-arg` bit. For example, locally I've been able to do the following:
```console
earthly +test-mutations --n_mutations=1000 --modules=src/parameters.ml
```
Just pointing this out in case it helps avoiding the duplication above.

dorranh:
The mutation workflow seems to be working so far. Manually triggered run results: https://github.com/tezos-checker/checker/runs/3610549532?check_suite_focus=true

gkaracha (commented):
Great that the mutation workflow seems to be working OK :ok_hand: If the only way to test the testnet deployment is on `master` I'd say there's no need to delay merging it. If that's what GH Actions wants.. :unamused: (For what it's worth, it was the same with the message bot; we did all the debugging on `master`.)

.github/workflows/mutate.yml
dorranh: Oh wow, I didn't know that syntax existed. Thanks!

Issue #248, opened 2021-09-15
List of mutations to process


gkaracha:
Until now we've addressed several mutations, but in a rather ad-hoc way. To be more systematic about it, we can crank up the number of mutations (to, say, a thousand) and focus on one file at a time. Here is a WIP list of mutations obtained this way that we can work on.

`earthly +test-mutations --n_mutations=1000 --modules=src/avl.ml`
---

- [ ] src/avl.ml           SWAP_FUNCTION        lt_int_int      -> eq_int_int      L285
- [ ] src/avl.ml           INTEGER_LITERAL      2n              -> 1n              L376
- [ ] src/avl.ml           INTEGER_LITERAL      0               -> -1              L277
- [ ] src/avl.ml           INTEGER_LITERAL      0n              -> 1n              L716
- [ ] src/avl.ml           INTEGER_LITERAL      1n              -> 0n              L71
- [ ] src/avl.ml           SWAP_FUNCTION        leq_int_int     -> geq_int_int     L287
- [ ] src/avl.ml           INTEGER_LITERAL      1n              -> 0n              L75
- [ ] src/avl.ml           SWAP_FUNCTION        leq_int_int     -> gt_int_int      L287
- [ ] src/avl.ml           SWAP_FUNCTION        leq_int_int     -> lt_int_int      L287
- [ ] src/avl.ml           INTEGER_LITERAL      0               -> 1               L277
- [ ] src/avl.ml           SWAP_FUNCTION        lt_int_int      -> leq_int_int     L285
- [ ] src/avl.ml           INTEGER_LITERAL      1n              -> 2n              L71
- [ ] src/avl.ml           SWAP_FUNCTION        leq_int_int     -> eq_int_int      L287
- [ ] src/avl.ml           SWAP_FUNCTION        lt_int_int      -> eq_int_int      L277
- [ ] src/avl.ml           SWAP_FUNCTION        lt_int_int      -> leq_int_int     L277

`earthly +test-mutations --n_mutations=1000 --modules=src/burrow.ml`
---

- [ ] src/burrow.ml        INTEGER_LITERAL      0               -> -1              L451

`earthly +test-mutations --n_mutations=1000 --modules=src/cfmm.ml`
---

Nothing here.

`earthly +test-mutations --n_mutations=1000 --modules=src/error.ml`
---

Unnecessary to mutate.

PR #253, opened 2021-09-20
Updates to the mutation script

gkaracha:
1. Extend `MUTATION_GROUPS` with more mutations. Until now we've
   had some entries for `fixedPoint.ml`, but other newtypes (e.g.,
   tok, kit, ..) did not have any rules.
2. Make `test_mutated_src` accept non-compiling mutations as tested.
   This silent recovery allows us to run the script on CI on more
   files without intervention.

### DIFF
--- a/scripts/mutate.py
+++ b/scripts/mutate.py
@@ -47,7 +47,26 @@ MUTATION_GROUPS = [
     # KIT
     {"kit_zero", "kit_one"},
     {"kit_add", "kit_sub", "kit_min", "kit_max"},
-    {"gt_kit_kit", "geq_kit_kit", "eq_kit_kit", "lt_kit_kit"},
+    {"geq_kit_kit", "lt_kit_kit", "gt_kit_kit", "eq_kit_kit"},
+    {"kit_of_fraction_ceil", "kit_of_fraction_floor"},
+    # CTEZ
+    {"ctez_add", "ctez_sub"},
+    {"eq_ctez_ctez", "lt_ctez_ctez", "gt_ctez_ctez"},
+    {"ctez_of_fraction_ceil", "ctez_of_fraction_floor"},
+    # LQT
+    {"lqt_add", "lqt_sub"},
+    {"lqt_zero", "lqt_one"},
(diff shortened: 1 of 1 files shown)


gkaracha:
@dorranh I noticed that `test_mutated_src` in `mutate.py` fails when the mutated source does not compile. Wouldn't it be better to accept the mutation as "tested" instead? Mutation rules for kit for example when mutating `kit.ml` might mutate the definition of a function, making the source ill-typed. Instead of having to change the mutation groups manually, I think it best to just move on. This way both locally and on CI we'll be able to execute the mutation script without intervention. What do you think?

dorranh:
@gkaracha, that's a good point. We're starting to bump up against the limits of the script's intended usage (catching some mutations vs exhaustively checking mutations). In principle I don't see an issue with skipping faulty mutations. I think that in the current implementation this attempted mutation will still count against the `n_mutations` you specify, but this might actually be desirable behavior since it will help avoid cases where the runtime starts to get long because faulty mutations are being attempted. To improve performance we might also want to consider recording failing mutations in memory so that they are only attempted once.

gkaracha:
You're absolutely right, we're past the originally intended usage. For now I would accept the mutations that led to an incompilable source, even if they still count against the specified `n_mutations`; this should allow me to progress with the batch-mutation I've been working on as part of #248 (and perhaps another round after that). Later we could improve the script even more, though I am not sure if it will ever become a priority, perfecting the script.

dorranh (approved):
Nice! It'll be good to test all of these extra modules, though I hope it doesn't add too much noise to the tests (i.e. that we don't have to dramatically increase `n_mutations` to continue getting enough legit test cases). However, there's not a simple way around that trade off which I can see, and I'm happy merging this :slightly_smiling_face:

gkaracha:
Good point :thinking: For what it's worth, it's this branch that I've been using locally to generate the data for #248 (currently containing 97 mutations to add tests for) with `n_mutations = 1000`. I guess after completing a first round of adding tests for the mutations in #248 we could try raising `n_mutations` even more and see what we get.

PR #254, opened 2021-09-21
Enable configuration of collateral token scaling factor

### LINKED ISSUE #251: Configurable scaling factor for collateral
#228 conveniently abstracted all collateral token logic into `tok.ml`, which is currently hard-wired to use the same scaling factor as tez. As we move towards FA2 collateral (#213) it would be good to make this value configurable as a part of the build so that the correct scaling factor can be easily specified for other tokens without having to modify the source code.

dorranh:
This PR enables the `tok` scaling factor to be configured at build time via a new `checker.yaml` config file. Instead of simply doing a `sed` on `tok.ml`, this PR uses a template engine (jinja2) to generate `tok.ml`. While this is a bit overkill for this specific issue, it provides the scaffolding for more complex code generation logic which will be required as we continue to work on #213 along with providing the tools needed to make system constants configurable. For example, using this approach we can easily generate modules for the various collateral and cfmm tokens for the different Checker variants outlined [here](https://github.com/tezos-checker/checker/issues/213#issuecomment-923758356). It also provides the Python scaffolding for porting over other build scripts (e.g `generate-entrypoints.rb`) in the future.

Closes #251


scripts/builder/checker_builder/templates/constants.ml.jinja
```diff
@@ -0,0 +1,7 @@
+(* Automatically generated by checker_builder. DO NOT EDIT DIRECTLY! *)
+
+open Common
+open FixedPoint
+open Tok
+
+(* TODO: Implement this template *)
```
gkaracha, line 7: I think this is not needed yet, right? In a separate PR we can generate the `constants.ml` file as well.

scripts/builder/checker_builder/config.py
```diff
@@ -0,0 +1,104 @@
...
+#     """Creates a dataclass field with a comment in its metadata"""
+#     return field(default=default, metadata={"comment": comment})
+# @dataclass
+# class SystemParameters:
+#     """Checker system parameters"""
+
+#     fminting: Decimal = commented(
+#         "Dimensionless. Factor used for setting the minting limit.",
+#         default=(Decimal(21) / Decimal(10)),
+#     )
```
gkaracha, line 37: Right, this one has to go as well. I am still thinking about the parameters actually, since their types can vary and I am not sure what's the best way to get them from a file (even though it would be nice for people deploying checker not having to look at the OCaml code at all). We would definitely need to avoid floating point numbers from python, and other things that could create inconsistencies. I guess passing strings would be the safest (though totally untyped, of course). :thinking: This is all for another PR of course.

scripts/builder/checker_builder/config.py
```diff
@@ -0,0 +1,104 @@
...
+
+
+@dataclass
+class CollateralTokenConfig:
+    scaling_factor: int = 1_000_000
+
+    @property
+    def decimal_digits(self) -> int:
+        return math.floor(math.log10(self.scaling_factor))
+
```
gkaracha, line 47: Hmmm, it's probably better to go the other way: get the decimal digits as an input, and compute the scaling factor by exponentiation. This way one cannot for example end up with inconsistencies such as
```
scaling factor = 1_234_567
decimal_digits = 6
```
because of `math.floor`. Actually, I think going through python's numbers might be a hazard and precision might vary. What if we used strings instead?

scripts/builder/checker_builder/config.py
dorranh: I think that we should* actually be pretty safe in this regard still. If someone specifies a decimal like `"2.1"` we can extract that as a fixedpoint (`Decimal`) in Python and convert it to a fraction so that it ultimately is rendered as a `ratio` in the generated source (i.e. the same way that they are currently implemented in `constants.ml`)

scripts/builder/checker_builder/config.py
dorranh: Thanks, @gkaracha! This might have been a misunderstanding on my end. Can the scaling factors be set to values other than multiples of ten? If not then I can definitely reverse this.

gkaracha: For most (all?) tokens as far as I know scaling factors are powers of 10 (so the name "_**decimal**_ digits" also implies that the scaling factor is a power of 10).

Our fixedpoint is the only exception to this rule I think, since it has a scaling factor of `18446744073709551616` (which is `2^64`), but that's an internal thing; coins as far as I know always use scaling factors that are powers of 10.

dorranh: Okay awesome, thanks for the clarification. I'll swap this around to take the decimal digits as its input.

PR #255, opened 2021-09-22
Implement the tez wrapper contract

gkaracha:
As described in https://github.com/tezos-checker/checker/issues/213#issuecomment-919838046. Currently WIP.

In addition to adding a draft of the tez wrapper contract, this PR performs some general changes as well.
* The contents of `fa2Interface.ml` are partitioned into three files:
  - `fa2Interface.ml` contains all user-facing FA2-conforming types (general purpose)
  - `fa2Ledger.ml` contains the definition of `fa2_state` and functions operating on it (general purpose)
  - `fa2Implementation.ml` contains the implementations of the three FA2 entrypoints (`fa2_run_balance_of`, `fa2_run_update_operators`, and `fa2_run_transfer`) (checker-specific).

  The reason for this partitioning is so that we can reuse the common parts (API in `fa2Interface.ml` and generic infrastructure in `fa2Ledger.ml`) for the tez wrapper contract, while allowing the tez wrapper contract to implement its own versions of the FA2 entrypoints (which are indeed different than those of checker).

  NOTE: All three modules are tested in `testFa2.ml`.

* Inline pragmas are added to three functions (`set_fa2_ledger_value`, `ledger_issue`, and `ledger_withdraw`) to avoid gas explosion. Ideally we would not inline any of these functions, but the order of definitions in ligo programs can have a (surprisingly) serious impact on gas costs. The partitioning above required some reordering of definitions which led to failures, hence the inline pragmas. This makes some sense to me since I expect the stack to be different, but such an effect on gas costs feels like a LIGO bug to me.


gkaracha:
Now it should be complete, though it's pretty rough around the edges. There are a few things I need to fix before undrafting the PR. I am especially bothered by how specialized `fa2Interface.ml` is to checker; I might end up refactoring it into two modules: the `fa2Interface` that is indeed shared among all FA2-contracts (so it can be reused by both Checker and TezWrapper) and the checker-specific parts (e.g., valid tokens being kit and lqt). Other things that are certainly missing: tests, fa2 metadata.

@dorranh could you also have a look and let me know if you see any red flags? I know it's still a draft but I could use a second pair of eyes :sweat_smile:

src/tezWrapper.ml
```diff
@@ -0,0 +1,279 @@
...
+let[@inline] withdraw (state: tez_wrapper_state) (amnt: Ligo.tez) : LigoOp.operation list * tez_wrapper_state =
+  (* 1. Ensure no tez given *)
+  let _ = ensure_no_tez_given () in
+  (* 2. Reduce the balance of the tez owner *)
+  let state_fa2_state = ledger_withdraw_tez_token (state.fa2_state, !Ligo.Tezos.sender, amnt) in
+  let state = { state with fa2_state = state_fa2_state; } in
+  (* 3. Create a vault, if it does not exist already, and update the map. *)
+  let op_opt, state, vault_address = find_vault_address state !Ligo.Tezos.sender in
+  (* 4. Instruct the vault to send the actual tez to the owner *)
+  let op = match (LigoOp.Tezos.get_entrypoint_opt "%vault_send_tez_to_contract" vault_address : (Ligo.tez * Ligo.address) Ligo.contract option) with
```
dorranh, line 246: I'm wondering if we want to send the tez to a contract or via a bare transaction here to enable withdrawals directly to user accounts? Not sure if there are any negative implications to doing this though

src/ligoOp.ml
```diff
@@ -81,8 +89,13 @@ module Tezos = struct
 
   let get_contract_opt address = Some (contract_of_address address)
 
-  let create_contract code delegate tez store =
-    let contract = CreateContract (code, delegate, tez, store) in
+  let burrow_create_contract code delegate tez store =
+    let contract = CreateBurrowContract (code, delegate, tez, store) in
+    let address = get_next_address () in
```
dorranh, line 94: > get_next_address()

Ah, I had always wondered how we get the new addresses prior to applying the operation :sweat_smile:

src/tezWrapper.ml
```diff
@@ -0,0 +1,279 @@
...
+    | Some c -> LigoOp.Tezos.tez_address_transaction (amnt, !Ligo.Tezos.sender) (Ligo.tez_from_literal "0mutez") c
+    | None -> (Ligo.failwith error_GetEntrypointOptFailureVaultSendTezToContract : LigoOp.operation) in
+  match op_opt with
+  | None -> ([op], state)
+  | Some origination -> ([origination; op], state)
+
+let[@inline] set_delegate (state: tez_wrapper_state) (kho: Ligo.key_hash option) : LigoOp.operation list * tez_wrapper_state =
```
dorranh, line 253: I'm not sure if we need to create the vault if it doesn't exist when calling this entrypoint, but it does at least make the behavior consistent with the FA2 entrypoints

src/tezWrapper.ml
```diff
@@ -0,0 +1,279 @@
...
+      (fun (((st, ops), tx): (tez_wrapper_state * LigoOp.operation list) * fa2_transfer) ->
+         let { from_ = from_; txs = txs; } = tx in
+
+         (* Origination of the from_ vault, if needed *)
+         let op_opt, st, from_vault_address = find_vault_address st from_ in
+         let ops = match op_opt with
+           | None -> ops (* vault is already originated *)
+           | Some origination -> (origination :: ops) in (* originate now *) (* NOTE: reversed! *)
+
+         (* TODO: Do we really need to restrict the to_ and from_ here? *)
```
dorranh, line 170: What I worry about with this is that we could end up causing vaults to be created for vaults if someone transfers tez-tokens to a vault address. I'm not sure if this causes vulnerabilities, but it feels a bit odd.

e.g.

> I transfer to address of Vault A -> No record exists in ledger or Vault A -> Vault for Vault A is originated and transfer is executed

At that point the tokens are lost to the vault since we can't issue transfers from Vault A directly, though I guess the end result is just that the tokens are lost (not much different than sending tokens to an unknown address)

dorranh (commented):
Nice! I had a look over the current draft. It looks good to me and is in line with the spec we had previously discussed. Just left a few general comments :slightly_smiling_face:

src/ligoOp.ml
gkaracha: Hehe, I know, it's not all that magical or pretty when you look at it from the inside :sweat_smile: The signature matches though the LIGO one; I have no idea how tezos implements this part internally :shrug:

src/tezWrapper.ml
gkaracha: That's a fair point, we could also fail without any problems. I mostly tried to make vault origination implicit (and eager) everywhere, so that, indeed, similarly to FA2, operations work fine even if an account is not created yet or if a user has zero tez, etc. I guess this approach is rather eager when it comes to vault creation, but I think that's actually nice: we never have to fail due to "vault non-existent" errors. I can't think of a downside at the moment but perhaps we shoudl revisit the approach.

src/tezWrapper.ml
gkaracha: Ah, I see what you mean. Yes, that could happen but indeed it's not much different than sending tokens to an unknown address. I'm sure that it cannot be used for an attack since we're using bigmaps (the performance of the tez wrapper will not degrade no matter how many vaults someone can create), but I guess it is a little weird. On the other hand, adding the check will raise gas costs, but people will still be able to create vaults for useless accounts if they want so the problem is not entirely gone. :thinking:

gkaracha: Hmmm, looking more into this, I think that implementing a check about whether an address belongs to a vault is more complicated than it seems at first glance: the big map is indexed by user address, so figuring out whether an address is a vault address cannot be done unless we keep track of them in a separate map. I added a check for `!Ligo.Tezos.self_address` for now, but we might have to reconvene about the vault check.

dorranh:
b'| Gas costs           | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |\n| ------------------- | ---------- | ---------- | ---------- |\n| touch               | 30396      | 30471      | 75         |\n| remove_liquidity    | 12556      | 12584      | 28         |\n| add_liquidity       | 12639      | 12654      | 15         |\n| buy_kit             | 12492      | 12502      | 10         |\n| sell_kit            | 11724      | 11734      | 10         |\n| withdraw_collateral | 11130      | 11133      | 3          |\n| create_burrow       | 6912       | 6915       | 3          |\n| activate_burrow     | 8918       | 8921       | 3          |\n| deposit_collateral  | 8898       | 8901       | 3          |\n| burn_kit            | 7151       | 7154       | 3          |\n| set_burrow_delegate | 9833       | 9836       | 3          |\n| mint_kit            | 6923       | 6926       | 3          |\n| deactivate_burrow   | 11276      | 11279      | 3          |\n| update_operators    | 5407       | 5410       | 3          |\n| transfer            | 6129       | 6131       | 2          |\n| touch_burrow        | 5919       | 5921       | 2          |\n\n| Entrypoint sizes              | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |\n| ----------------------------- | ---------- | ---------- | ---------- |\n| add_liquidity                 | 1967       | 1833       | -134       |\n| remove_liquidity              | 2038       | 1953       | -85        |\n| touch                         | 56590      | 56661      | 71         |\n| buy_kit                       | 1376       | 1350       | -26        |\n| sell_kit                      | 1372       | 1346       | -26        |\n| liquidation_auction_place_bid | 2229       | 2206       | -23        |\n| touch_liquidation_slices      | 14482      | 14465      | -17        |\n\n| Test coverage        | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |\n| -------------------- | ---------- | ---------- | ---------- |\n| fa2Implementation.ml | None       | 100        | 100        |\n| fa2Interface.ml      | 100        | 0          | -100       |\n| fa2Ledger.ml         | None       | 100        | 100        |\n| common.ml            | 100        | 97.87      | -2.1299999999999955 |\n| checker.ml           | 90.18      | 90.1       | -0.0800000000000125 |\n| TOTAL                | 93.93      | 93.87      | -0.060000000000002274 |\n\n' 

[hiddencomment]: bot:ed5c6244284ce7254ab9ba9d1812256a68d0fed8:7e6aa28ada58b186e58db0f76394f276fe76432a

dorranh:
| Gas costs           | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |
| ------------------- | ---------- | ---------- | ---------- |
| touch               | 30396      | 30471      | 75         |
| remove_liquidity    | 12556      | 12584      | 28         |
| add_liquidity       | 12639      | 12654      | 15         |
| buy_kit             | 12492      | 12502      | 10         |
| sell_kit            | 11724      | 11734      | 10         |
| update_operators    | 5407       | 5410       | 3          |
| set_burrow_delegate | 9833       | 9836       | 3          |
| withdraw_collateral | 11130      | 11133      | 3          |
| burn_kit            | 7151       | 7154       | 3          |
| create_burrow       | 6912       | 6915       | 3          |
| mint_kit            | 6923       | 6926       | 3          |
| deposit_collateral  | 8898       | 8901       | 3          |
| activate_burrow     | 8918       | 8921       | 3          |
| deactivate_burrow   | 11276      | 11279      | 3          |
| touch_burrow        | 5919       | 5921       | 2          |
| transfer            | 6129       | 6131       | 2          |

| Entrypoint sizes              | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |
| ----------------------------- | ---------- | ---------- | ---------- |
| add_liquidity                 | 1967       | 1833       | -134       |
| remove_liquidity              | 2038       | 1953       | -85        |
| touch                         | 56590      | 56661      | 71         |
| sell_kit                      | 1372       | 1346       | -26        |
| buy_kit                       | 1376       | 1350       | -26        |
| liquidation_auction_place_bid | 2229       | 2206       | -23        |
| touch_liquidation_slices      | 14482      | 14465      | -17        |

| Test coverage        | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |
| -------------------- | ---------- | ---------- | ---------- |
| fa2Implementation.ml | None       | 100        | 100        |
| fa2Ledger.ml         | None       | 100        | 100        |
| fa2Interface.ml      | 100        | 0          | -100       |
| common.ml            | 100        | 97.87      | -2.1299999999999955 |
| checker.ml           | 90.18      | 90.1       | -0.0800000000000125 |
| TOTAL                | 93.93      | 93.87      | -0.060000000000002274 |

 

[hiddencomment]: bot:ed5c6244284ce7254ab9ba9d1812256a68d0fed8:7e6aa28ada58b186e58db0f76394f276fe76432a

dorranh:
| Gas costs           | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |
| ------------------- | ---------- | ---------- | ---------- |
| touch               | 30396      | 30471      | 75         |
| remove_liquidity    | 12556      | 12584      | 28         |
| add_liquidity       | 12639      | 12654      | 15         |
| sell_kit            | 11724      | 11734      | 10         |
| buy_kit             | 12492      | 12502      | 10         |
| create_burrow       | 6912       | 6915       | 3          |
| activate_burrow     | 8918       | 8921       | 3          |
| burn_kit            | 7151       | 7154       | 3          |
| deactivate_burrow   | 11276      | 11279      | 3          |
| set_burrow_delegate | 9833       | 9836       | 3          |
| deposit_collateral  | 8898       | 8901       | 3          |
| mint_kit            | 6923       | 6926       | 3          |
| withdraw_collateral | 11130      | 11133      | 3          |
| update_operators    | 5407       | 5410       | 3          |
| transfer            | 6129       | 6131       | 2          |
| touch_burrow        | 5919       | 5921       | 2          |

| Entrypoint sizes              | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |
| ----------------------------- | ---------- | ---------- | ---------- |
| add_liquidity                 | 1967       | 1833       | -134       |
| remove_liquidity              | 2038       | 1953       | -85        |
| touch                         | 56590      | 56661      | 71         |
| sell_kit                      | 1372       | 1346       | -26        |
| buy_kit                       | 1376       | 1350       | -26        |
| liquidation_auction_place_bid | 2229       | 2206       | -23        |
| touch_liquidation_slices      | 14482      | 14465      | -17        |

| Test coverage        | ed5c6244284ce7254ab9ba9d1812256a68d0fed8 | 7e6aa28ada58b186e58db0f76394f276fe76432a | Diff       |
| -------------------- | ---------- | ---------- | ---------- |
| fa2Implementation.ml | None       | 100        | 100        |
| fa2Ledger.ml         | None       | 100        | 100        |
| fa2Interface.ml      | 100        | 0          | -100       |
| common.ml            | 100        | 97.87      | -2.1299999999999955 |
| checker.ml           | 90.18      | 90.1       | -0.0800000000000125 |
| TOTAL                | 93.93      | 93.87      | -0.060000000000002274 |

 

[hiddencomment]: bot:ed5c6244284ce7254ab9ba9d1812256a68d0fed8:7e6aa28ada58b186e58db0f76394f276fe76432a

src/tezWrapper.ml
dorranh: Ah true, I hadn't thought about that. Given that we're trying to avoid gas costs I agree that it might make more sense to prioritize that over user-friendliness here since allowing users to create extra vault contracts shouldn't affect the functionality of the wrapper.

dorranh (approved):
Just one open question [here](https://github.com/tezos-checker/checker/pull/255#discussion_r715045100), but otherwise LGTM! We should add some extra tests for the wrapper contract, but I'm happy creating a separate issue for this if you prefer.

gkaracha:
> We should add some extra tests for the wrapper contract, but I'm happy creating a separate issue for this if you prefer.

Yeah, totally. I didn't add any since now it's not called at all, but after #213 is completed we should be able to add some meaningful tests.

gkaracha:
Ah, the message bot is back to work, thanks @dorranh!

dorranh:
> Ah, the message bot is back to work, thanks @dorranh!

It's alive! :scientist:

Issue #171, opened 2021-06-25
Generalization


murbard:
Minor generalizations of Checker greatly increase its use cases. Broadly speaking, the following aspects should be modular:

1. allow FA1.2 / FA2 token as collateral for burrows instead of tez only. A single form of collateral is still accepted for a given checker instance, but different instances can have different collateral
2. the control formula that sets the drift derivative as a function of the target should be pluggable, in some cases, a continuous function might make sense as opposed to a bang-bang control
3. We can distinguish two big classes of contracts: those that attempt to replicate an index provided by an oracle (index-based), and those that attempt to replicate an existing token on the chain (token-based)
   1. For index-based we use the index, the ctez / tez cfmm price, and the ctez / kit cfmm price, with the ctez / kit cfmm being subsidized by the checker contract.
   2. For token-based, the setup is typically a bit different. Suppose we want to replicate USDS using tzBTC as collateral. In that case, we would want to derive prices from 3 cfmms, USDS / cUSDS, tzBTC / ctez and USDS / ctez, with the USDs / cUSDS pair being subsidized by the checker contract.

gkaracha:
I'm closing this issue, since we've been dealing with each part of it in separate issues (part 1 is tracked in #213, part 2 was discussed in #261 (and finally addressed in #283), and part 3 is tracked in #263).

PR #256, opened 2021-09-23
Expose system constants in build configuration

dorranh:
This PR turns the contents of `constants.ml` inside-out, moving them to the `checker.yaml` file and generating `constants.ml` at build-time. This allows for easier fine tuning of constants when building Checker, which will be required as we continue to generalize the contract (#171). 

Due to the tight conditions in our current test suite, most (hopefully all) changes to the constants in `checker.yaml` will break tests in the OCaml test suite, and this is noted in the README section on configuration. However, one is still able to build and deploy the contract with the modified constants, allowing for easy experimentation. The constants in the `checker.yaml` file checker into the repo are set to the old values in `constant.ml` and are meant to serve as a "source of truth" for all CI and local builds/tests.


scripts/builder/checker_builder/test/test_config.py
```diff
@@ -1,6 +1,88 @@
...
+
+    input = Ratio(-3, 997)
+    field = RatioField()
+    assert field.deserialize(field.serialize("ratio", obj={"ratio": input})) == input
+
+
+def test_deserialize_positive_ratio():
+    from checker_builder.config import PositiveRatioField, Ratio
+
+    assert PositiveRatioField().deserialize("-1/-9") == Ratio(-1, -9)
```
gkaracha, line 45: Hmmm, actually now that I look at this, we don't want to allow ratios with a negative denominator (i.e., this will fail on the OCaml side). In general we should have `den > 0` and get the sign from the numerator.

Earthfile
```diff
@@ -161,10 +161,15 @@ generate-code:
     # Ensure that the generated modules obey formatting rules:
     RUN opam exec -- ocp-indent -i ./src/*
 
+    # TODO: Find a way to drop all generated code in one directory
+    # so that we can extract it without having to explicitely name
```
gkaracha, line 165:
```suggestion
    # so that we can extract it without having to explicitly name
```

gkaracha (commented):
Nice work! :ok_hand: It looks like a lot of boilerplate, but I guess a lot of it comes from the checks on the python side, which are good to have. Left a few comments, but I might address a few of them myself in the upcoming days.

gkaracha (approved):
I think this one is good to go, I'll go ahead and merge it :slightly_smiling_face:

PR #257, opened 2021-09-28
Restore the entrypoint-sizes.json artifact

gkaracha:
Due to a typo the artifact has been just an empty file for some time now.


gkaracha:
Indeed this was it; see the (now populated) artifact here: https://github.com/tezos-checker/checker/runs/3732545417.

gkaracha:
Unless there are other issues, restoring this artifact should revive the message bot, which has been silent for a while.

gkaracha:
Actually, now that I look at it, `.github/workflows/comment.yml` is still invoking nix so there is more work to do here for the bot to be revived.

PR #262, opened 2021-10-01
Fix comment bot

dorranh:
Fixes the currently broken comment bot. Along the way I ended up porting the comment script to Python to allow it to be run locally without adding nodejs as a project dependency.

The updated script also has a slightly modified behavior compared to the previous bot. Previously only one comment would be created per PR, with subsequent commits not being reflected in the comment. The updated script will edit an existing gas 
 cost comment in-place on each run.

Example manual run: https://github.com/tezos-checker/checker/runs/3788833928?check_suite_focus=true


gkaracha (approved):
Sweet! Feel free to merge!

Issue #267, opened 2021-10-07
End to end tests for tez wrapper contract


dorranh:
Now that we have a separate tez wrapper contract, it would be good to implement e2e tests similar to those we have for Checker.

PR #268, opened 2021-10-07
Make roundabout calls to get_entrypoint_opt in TezWrapper

### LINKED ISSUE #266: Re-factor entrypoints which create tez vaults
As we discussed today, entrypoints in the FA2 wrapper for tez which generate vaults have an issue where despite generating valid operations, calls to them will fail due to how we call `get_entrypoint_opt`. 

More specifically, we currently call `get_entrypoint_opt` within the entrypoint to generate internal operations which interact with vaults, but since the origination hasn't been executed yet this method always fails. Refactoring these entrypoints (`deposit`, `withdraw`, and `transfer`) into two entrypoints which can be called before and after the origination should fix this issue.

gkaracha:
Fixes #266.

In order to avoid calling `get_entrypoint_opt` on a vault before it has been originated, the calls to
```ocaml
get_entrypoint_opt "%X" vault_address
```
are replaced by internal calls
```ocaml
get_entrypoint_opt "%call_X" Tezos.self_address
```
—each of the internal calls calling `get_entrypoint_opt "%X" vault_address` internally—so that the call to `get_entrypoint_opt` is executed _*after*_ the origination of the vault has happened (this technique works because it ensures that the origination and the entrypoint access are not part of the same operation).

### DIFF
--- a/src/tezWrapper.ml
+++ b/src/tezWrapper.ml
@@ -304,7 +304,7 @@ let[@inline] set_delegate (state: tez_wrapper_state) (kho: Ligo.key_hash option)
 (*****************************************************************************)
 
 let[@inline] call_vault_receive_tez (state: tez_wrapper_state) (vault_address, amnt : Ligo.address * Ligo.tez) : LigoOp.operation list * tez_wrapper_state =
-  if Ligo.Tezos.sender <> Ligo.Tezos.self_address then
+  if !Ligo.Tezos.sender <> !Ligo.Tezos.self_address then
     (Ligo.failwith error_UnauthorisedCaller : LigoOp.operation list * tez_wrapper_state)
   else
     let op = match (LigoOp.Tezos.get_entrypoint_opt "%vault_receive_tez" vault_address : unit Ligo.contract option) with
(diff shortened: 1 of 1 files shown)


src/tezWrapper.ml
```diff
@@ -287,14 +291,45 @@ let[@inline] set_delegate (state: tez_wrapper_state) (kho: Ligo.key_hash option)
...
+  if Ligo.Tezos.sender <> Ligo.Tezos.self_address then
+    (Ligo.failwith error_UnauthorisedCaller : LigoOp.operation list * tez_wrapper_state)
+  else
+    let op = match (LigoOp.Tezos.get_entrypoint_opt "%vault_send_tez_to_contract" vault_address : (Ligo.tez * Ligo.address) Ligo.contract option) with
+      | Some c -> LigoOp.Tezos.tez_address_transaction (amnt, recipient) (Ligo.tez_from_literal "0mutez") c
+      | None -> (Ligo.failwith error_GetEntrypointOptFailureVaultSendTezToContract : LigoOp.operation) in
+    ([op], state)
+
+let[@inline] call_vault_set_delegate (state: tez_wrapper_state) (vault_address, kho : Ligo.address * Ligo.key_hash option) : LigoOp.operation list * tez_wrapper_state =
+  if Ligo.Tezos.sender <> Ligo.Tezos.self_address then
```
dorranh, line 325: :+1:

dorranh (approved):
Fastest patches in the west :sunglasses: 

Nice, thanks for preparing this @gkaracha! It looks good as far as I can tell. I'm happy either merging this now then working on #267 or rebasing this branch on the one for that issue to be extra safe - as you prefer.

src/tezWrapper.ml
dorranh: Ah, you have sharper eyes than me today. Nice catch and sorry I didn't notice that in the review :grimacing:

gkaracha: Ha, no worries! Strictly-speaking this is not a bug; the final michelson will be correct. It's just a weakness of our compatibility layer, what can you do! :shrug:

dorranh:
Deployed to granadanet and manually tested `deposit`, `withdraw`, and `transfer`. Seems to be working to me. 

https://better-call.dev/granadanet/KT1FkLkNWDqratPAQueL5YRtkyGTgP1mDn6L/operations

gkaracha:
Awesome! I'll go ahead and merge it then. The developing and testing of this patch was indeed pretty fast; go us :sunglasses:

PR #269, opened 2021-10-08
Switch to FA2 for collateral, exclusively

gkaracha:
First part of #213: integration of the tez wrapper into checker's workflow (for details see section "Changes for Checker" in https://github.com/tezos-checker/checker/issues/213#issuecomment-919838046). In essence this PR makes checker deal with FA2 exclusively. To keep CI green, usage of the tez wrapper is hardwired; adding the switch to be able to change the collateral type should be part of a separate PR. This PR is already pretty big.

Other changes:
* Removed the restrictions on `from_` and `to_` in `TezWrapper.transfer` to save on gas costs (they were optional anyway).
* Fixed a bug in `TezWrapper.transfer` (transfer of zero tez should not fail due to non-originated vaults).
* Made `tezWrapper.ml` more verbose to significantly reduce its gas costs (see bot message below).

`TezWrapper.transfer` is still pretty expensive which makes Checker's entrypoints expensive too. This we can improve incrementally after merging this PR.

### DIFF
--- a/client/checker_client/cli.py
+++ b/client/checker_client/cli.py
@@ -176,6 +176,7 @@ def deploy(config: Config, address=None, port=None, key=None):
     show_default=True,
 )
 @click.option("--oracle", type=str, help="Oracle contract address")
+@click.option("--tez_wrapper", type=str, help="TezWrapper contract address")
 @click.option("--ctez", type=str, help="ctez contract address")
 @click.option(
     "--token-metadata",
@@ -183,13 +184,17 @@ def deploy(config: Config, address=None, port=None, key=None):
     help="optional JSON file containing the TZIP-12 token_metadata.",
 )
 @click.pass_obj
-def checker(config: Config, checker_dir, oracle, ctez, token_metadata):
+def checker(config: Config, checker_dir, oracle, tez_wrapper, ctez, token_metadata):
     """
     Deploy checker. Requires addresses for oracle and ctez contracts.
     """
     if not config.oracle_address and not oracle:
         raise ValueError(
             "Oracle address was neither specified in the CLI config or provided as an argument."
+        )
+    if not config.tez_wrapper and not tez_wrapper:
+        raise ValueError(
+            "TezWrapper address was neither specified in the CLI config or provided as an argument."
         )
     if not config.ctez_address and not ctez:
         raise ValueError(
@@ -206,6 +211,7 @@ def checker(config: Config, checker_dir, oracle, ctez, token_metadata):
     checker = checker_lib.deploy_checker(
         client,
         checker_dir,
+        tez_wrapper=config.tez_wrapper_address,
         oracle=config.oracle_address,
         ctez=config.ctez_address,
         ttl=_patch_operation_ttl(config),
--- a/client/checker_client/cli.py
+++ b/client/checker_client/cli.py
@@ -190,15 +190,15 @@ def checker(config: Config, checker_dir, oracle, tez_wrapper, ctez, token_metada
     """
     if not config.oracle_address and not oracle:
         raise ValueError(
-            "Oracle address was neither specified in the CLI config or provided as an argument."
+            "Oracle address was neither specified in the CLI config nor provided as an argument."
         )
     if not config.tez_wrapper and not tez_wrapper:
         raise ValueError(
-            "TezWrapper address was neither specified in the CLI config or provided as an argument."
+            "TezWrapper address was neither specified in the CLI config nor provided as an argument."
         )
     if not config.ctez_address and not ctez:
         raise ValueError(
-            "ctez address was neither specified in the CLI config or provided as an argument."
+            "ctez address was neither specified in the CLI config nor provided as an argument."
         )
     if oracle:
         config.oracle_address = oracle
--- a/client/checker_client/cli.py
+++ b/client/checker_client/cli.py
@@ -211,8 +211,8 @@ def checker(config: Config, checker_dir, oracle, tez_wrapper, ctez, token_metada
     checker = checker_lib.deploy_checker(
         client,
         checker_dir,
-        tez_wrapper=config.tez_wrapper_address,
         oracle=config.oracle_address,
+        tez_wrapper=config.tez_wrapper_address,
         ctez=config.ctez_address,
         ttl=_patch_operation_ttl(config),
         token_metadata_file=token_metadata,
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -323,8 +323,8 @@ class E2ETest(SandboxedTestCase):
         checker = deploy_checker(
             self.client,
             checker_dir=CHECKER_DIR,
-            tez_wrapper=tez_wrapper.context.address,
             oracle=oracle.context.address,
+            tez_wrapper=tez_wrapper.context.address,
             ctez=ctez["fa12_ctez"].context.address,
             ttl=MAX_OPERATIONS_TTL,
         )
@@ -645,8 +645,8 @@ class LiquidationsStressTest(SandboxedTestCase):
         checker = deploy_checker(
             self.client,
             checker_dir=CHECKER_DIR,
-            tez_wrapper=tez_wrapper.context.address,
             oracle=oracle.context.address,
+            tez_wrapper=tez_wrapper.context.address,
             ctez=ctez["fa12_ctez"].context.address,
             ttl=MAX_OPERATIONS_TTL,
         )
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -828,7 +828,7 @@ class LiquidationsStressTest(SandboxedTestCase):
             flattened_cancel_ops += [op1, op2]
         call_bulk(
             flattened_cancel_ops,
-            batch_size=170,
+            batch_size=100,
             profiler=cancel_liquidation_slice_profiler,
         )
 
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -24,9 +24,10 @@ WRITE_GAS_COSTS = os.getenv("WRITE_GAS_COSTS")
 # tests
 WRITE_GAS_PROFILES = os.getenv("WRITE_GAS_PROFILES")
 
-# FA2 token id for tokens issued by the tez wrapper contract
-# FIXME: Find a way to get this value programatically
-TEZ_TOKEN_ID = 2
+# FIXME: How to determine this field programmatically?  It must be equal to
+# tezWrapper.tez_token_id when collateral=tez or the same as the one used by
+# the FA2 we use as collateral.
+COLLATERAL_TOKEN_ID = 2
 
 
 class SandboxedTestCase(unittest.TestCase):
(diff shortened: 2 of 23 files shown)


client/checker_client/cli.py
```diff
@@ -206,6 +206,7 @@ def checker(config: Config, checker_dir, oracle, ctez, token_metadata):
     checker = checker_lib.deploy_checker(
         client,
         checker_dir,
+        tez_wrapper=config.tez_wrapper_address,
```
dorranh, line 209: We should probably add an additional check similar to line 194 to show a friendly error message if the wrapper address isn't specified

https://github.com/tezos-checker/checker/pull/269/files#diff-65a5c024a0df2eb7077947147e98a51765d31b858c3a7c145fbd4d773cfc0063R194

e2e/main.py
```diff
@@ -302,11 +310,25 @@ def call_checker_endpoint(
...
             return ret
 
+        def get_tez_tokens_and_make_checker_an_operator(amnt):
+            call_endpoint(tez_wrapper, "deposit", (), amount=amnt)
+            update_operators = [
+                {
+                    "add_operator": {
+                        "owner": account,  # bob?
+                        "operator": checker.context.address,  # checker
+                        "token_id": 3,  # FIXME: How to determine this field?
```
dorranh, line 320: This is a FIXME in a few places actually. For kit I believe that we can pull it from `functions.json`, but we aren't currently producing anything like that for the wrapper.

src/checker.ml
```diff
@@ -444,9 +510,9 @@ let touch_liquidation_slice
   let state_fa2_state = new_state_fa2_state in
 
   (* Signal the burrow to send the tez to checker. *)
```
dorranh, line 512:
```suggestion
  (* Signal the burrow to send the collateral to checker. *)
```

e2e/main.py
```diff
@@ -512,9 +545,23 @@ def call_bulk(bulks, *, batch_size, profiler=general_gas_profiler):
                 profile = profiler(self.client, checker, batch)
                 self.gas_profiles = merge_gas_profiles(self.gas_profiles, profile)
 
+        def get_tez_tokens_and_make_checker_an_operator(checker, amnt):
+            call_endpoint(checker, tez_wrapper, "deposit", (), amount=amnt)
```
dorranh, line 549:
```suggestion
            call_endpoint(tez_wrapper, "deposit", None, amount=amnt)
```

e2e/main.py
gkaracha: Yeah, that's a good point. I added a
```python
TEZ_TOKEN_ID = 2
```
at the top of the file like you had previously done, but eventually we should have a single source of truth, ideally the checker yaml file.

src/burrowTypes.ml
```diff
@@ -3,15 +3,14 @@ open LiquidationAuctionTypes
 
 type burrow_storage =
   { checker_address: Ligo.address;
-    burrow_id: burrow_id
+    burrow_id: burrow_id;
+    collateral_fa2: Ligo.address;
```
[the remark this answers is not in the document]
gkaracha: > I'm wondering, do we still use `burrow_id` anywhere in the burrow contract? If not, perhaps we could drop this from its storage?

Ah, good catch, you're absolutely right, not anymore we don't! It was needed for checker to be able to ensure that `%receive_slice_from_burrow` is only called from burrows, but that entrypoint is not removed. I will remove it :pray:

src/checker.ml
```diff
@@ -126,43 +131,82 @@ let[@inline] ensure_valid_leaf_ptr (mem: mem) (leaf_ptr: leaf_ptr) : unit =
...
-             let op = match (LigoOp.Tezos.get_entrypoint_opt "%receive_slice_from_burrow" !Ligo.Tezos.sender : burrow_id Ligo.contract option) with
-               | Some c -> LigoOp.Tezos.address_nat_transaction storage.burrow_id tz c
-               | None -> (Ligo.failwith (Ligo.int_from_literal "-3") : LigoOp.operation) in
+           | BurrowTransfer p ->
+             let (addr, amnt) = p in
+             let transfer =
+               { from_ = !Ligo.Tezos.self_address; (* from: FA2 account of burrow contract *)
+                 txs = [
+                   { to_ = addr;                   (* to: FA2 account of the given address *)
+                     token_id = Ligo.nat_from_literal "3n"; (* FIXME: How to determine this field? *)
```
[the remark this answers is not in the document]
gkaracha: > I think the collateral `token_id` will need to either a.) get specified in the `checker.yaml` config file or b.) provided with a call to sealContract(). Any thoughts?

Yes, absolutely agree. a.) would be the best.

src/checker.ml
```diff
@@ -173,15 +217,26 @@ let entrypoint_touch_burrow (state, burrow_id: checker * burrow_id) : LigoOp.ope
...
+  let _ = ensure_no_tez_given () in
   let burrow_id = (!Ligo.Tezos.sender, burrow_no) in
   let burrow = find_burrow state.burrows burrow_id in
   let _ = ensure_burrow_has_no_unclaimed_slices state.liquidation_auctions burrow_id in
-  let burrow = burrow_deposit_collateral state.parameters (tok_of_tez !Ligo.Tezos.amount) burrow in
-  let op = match (LigoOp.Tezos.get_entrypoint_opt "%burrowStoreTez" (burrow_address burrow) : unit Ligo.contract option) with
-    | Some c -> LigoOp.Tezos.unit_transaction () !Ligo.Tezos.amount c
-    | None -> (Ligo.failwith error_GetEntrypointOptFailureBurrowStoreTez : LigoOp.operation) in
+  let burrow = burrow_deposit_collateral state.parameters tok burrow in
+  let transfer =
```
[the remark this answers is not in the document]
gkaracha: I tried that actually, but then said helper function needs to take two arguments of type `address` and we don't have named arguments, so I thought it might be a bad idea (because we could accidentally switch `from` and `to` and it would all compile just fine). Totally agree that it's verbose though, I'd also like to abstract over the pattern too :confused:

e2e/main.py
```diff
@@ -531,9 +581,20 @@ def call_bulk(bulks, *, batch_size, profiler=general_gas_profiler):
...
+        #     ],
+        #     batch_size=350,
+        # )
         call_bulk(
             [
-                checker.create_burrow((burrow_id, None)).with_amount(100_000_000)
+                checker.create_burrow((burrow_id, None, 100_000_000))
                 for burrow_id in burrows
             ],
             batch_size=350,
```
dorranh, line 600: Just FYI - Based on the latest CI run, it looks like this batch size is too large (in terms of gas costs) now that we emit extra internal operations. I had this pretty tightly tuned before since it affects the total runtime of the test suite, but feel free to reduce the batch size as needed. https://github.com/tezos-checker/checker/runs/3869898775#step:6:7246

src/checker.ml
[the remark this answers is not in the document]
dorranh: That's too bad. In that case no worries - it's also not so bad having to explicitly specify the record. The only thing which we might still want to do is use an inlined variable for the token id to avoid accidentally writing the wrong value.

e2e/main.py
dorranh: PS: this might be an issue for any of the other lines which call `call_bulk()` if they include calls to entrypoints which now interact with the FA2 contract internally.

gkaracha: Ah, thanks! :pray: I started looking into batching now, I really hope it's the last obstacle for CI.

src/checker.ml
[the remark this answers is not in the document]
gkaracha: Now that you opened #271 I'll mark this comment as resolved.

src/checker.ml
```diff
@@ -126,43 +131,82 @@ let[@inline] ensure_valid_leaf_ptr (mem: mem) (leaf_ptr: leaf_ptr) : unit =
   | Some (Leaf _) -> ()
   | _ -> Ligo.failwith error_InvalidLeafPtr
 
-let[@inline] entrypoint_create_burrow (state, (burrow_no, delegate_opt): checker * (Ligo.nat * Ligo.key_hash option)) =
+(* TODO: if no tez is allowed to be given anywhere, it might be a better idea
```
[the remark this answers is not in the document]
gkaracha: So, about this, I really want to try the sharing, but it will be quite a change for a few reasons:
* it affects _all_ entrypoints
* we have many tests that check per entrypoint that they fail if any tez is given which will have to be patched
* by moving code from entrypoint code (serialized) to checkerMain (non-serialized) gas costs might go up

So, I suggest that I open a separate PR after we've merged this one. What do you think @dorranh?

e2e/main.py
gkaracha: Yay, @dorranh I think I managed to get the e2e tests to pass! :smile:

dorranh: Huzzah!

src/checker.ml
```diff
@@ -126,43 +131,82 @@ let[@inline] ensure_valid_leaf_ptr (mem: mem) (leaf_ptr: leaf_ptr) : unit =
...
-           | BurrowSendTezTo p ->
-             let (tez, addr) = p in
-             let op = match (LigoOp.Tezos.get_contract_opt addr : unit Ligo.contract option) with
-               | Some c -> LigoOp.Tezos.unit_transaction () tez c
-               | None -> (Ligo.failwith (Ligo.int_from_literal "-2") : LigoOp.operation) in
+             (* TODO: this deviates slightly from the design in the issue;
+              * I think the case analysis should happen in burrow code, and
+              * not in checker code. *)
+             (* TODO: Also, I am not happy that an FA2 contract could "pretend"
+              * to have a set_delegate entrypoint and succeed here. *)
```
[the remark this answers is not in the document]
gkaracha: I guess it's not that big of an issue after all :thinking:


I considered whether delegation should happen from the TezWrapper side (would also significantly lower the gas costs, not having to go through checker), but that can't happen easily: (a) the burrow contracts only listen to checker, and (b) only checker knows their address.

I considered exposing the burrow address for a user via an offline view to address (b), and changing the current logic to circumvent (a). There's still the issue of how much data would we need to keep in the vault's storage though to change the permission checking.

Not sure if any of that will work, but it might be worth considering it as an improvement after we merge this PR.

src/tezWrapper.ml
```diff
@@ -87,16 +86,19 @@ type tez_wrapper_params =
...
-let[@inline] find_vault_address (state: tez_wrapper_state) (user: Ligo.address) : LigoOp.operation option * tez_wrapper_state * Ligo.address =
-  match Ligo.Big_map.find_opt user state.vaults with
-  | Some vault_address -> (None: LigoOp.operation option), state, vault_address
+let[@inline] find_vault_address_append (vaults: vault_map) (user: Ligo.address) (ops: LigoOp.operation list) : vault_found * LigoOp.operation list * vault_map * Ligo.address =
+  match Ligo.Big_map.find_opt user vaults with
+  | Some vault_address -> VaultFound, ops, vaults, vault_address
   | None ->
     let op, vault_address = originate_vault !Ligo.Tezos.self_address in
-    Some op, {state with vaults = Ligo.Big_map.update user (Some vault_address) state.vaults}, vault_address
+    VaultNotFound, (op :: ops), (Ligo.Big_map.update user (Some vault_address) vaults), vault_address
```
dorranh, line 101: This is not an issue for this PR, but more out of curiosity - I wonder if there is a performance (gas cost) overhead associated with using a custom `vault_found` type here versus a boolean.

dorranh (approved):
Nice! I had a look at the latest changes, and they look good to me. We'll definitely need to add more tests once we're done with all of the surgery (as well as look at gas costs like you mentioned), but I'm happy with landing these changes.

Thanks for all of your hard work on this! :slightly_smiling_face:

src/tezWrapper.ml
gkaracha: Ah, that's a good question. There is no extra gas cost, since both types are represented in Michelson something like `or unit unit`. It's actually pretty nice, being able to use a domain specific type here without any runtime cost :smile:

PR #270, opened 2021-10-08
Add end to end tests for tez wrapper

### LINKED ISSUE #267: End to end tests for tez wrapper contract

dorranh:
Adds end to end tests which call all of the tez wrapper's entrypoints. In order to avoid further changes to our bot, etc. I updated the end to end tests to merge their gas costs, adding an identifier before each entrypoint (e.g. `checker%mint_kit`) to distinguish between overlapping entrypoint names. This might cause the github actions bot to produce some weird diff messages for this PR.

Closes #267

### DIFF
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -566,7 +566,7 @@ class TezWrapperTest(SandboxedTestCase):
             contract=wrapper_alice,
         )
 
-        # Send some kit back to the main test account
+        # Send some tez tokens back to the main test account
         call_endpoint("transfer", single_fa2_transfer(account_alice, account, 80))
         assert_wrapped_tez_balance(wrapper, account, 1_999_890)
         # `balance_of` requires a contract callback when executing on-chain. To make tests


dorranh:
Well, I got it fixed.  Turns out I hadn't pulled the bug fix for the wrapper from last week before creating this branch :facepalm:

e2e/main.py
```diff
@@ -45,7 +45,18 @@ def assert_kit_balance(checker: ContractInterface, address: str, expected_kit: i
...
     kit_token_id = 0
     kit_balance = checker.metadata.getBalance((address, kit_token_id)).storage_view()
-    assert expected_kit == kit_balance
+    if expected_kit != kit_balance:
+        raise AssertionError(f"Expected {expected_kit} but got {kit_balance}")
+
+
+def assert_wrapped_tez_balance(wrapper: ContractInterface, address: str, expected: int):
+    fa2_balance_of = {
+        "requests": [{"owner": address, "token_id": 2}],
```
gkaracha, line 54: Ah, this token id should also come from `tezWrapper.ml` actually (or, even better, from `checker.yml`). Let's discuss this a little more before merging, because we need access to the `tez_token_id` in #269 as well. It might be best to open a small PR first, where this field moves to `checker.yaml` and we also bring it to the python world similarly to `kit_token_id`, and then rebase this PR and #269 on top of it. I'm afraid that one of the values will get out of sync (though #269 is probably much more in risk than this pr).

e2e/main.py
```diff
@@ -436,16 +471,117 @@ def call_checker_endpoint(
...
+                getattr(contract, name)(param)
+                .with_amount(amount)
+                .as_transaction()
+                .autofill(ttl=MAX_OPERATIONS_TTL)
+                .sign(),
+            )
+            gas_costs[f"tezWrapper%{name}"] = int(ret["contents"][0]["gas_limit"])
+            return ret
+
+        def single_fa2_transfer(sender: str, recipient: str, amount: int, token_id=2):
```
gkaracha, line 513: Also here it's a little scary having the `2` be inlined.

e2e/main.py
```diff
@@ -436,16 +471,117 @@ def call_checker_endpoint(
...
+            },
+        ]
+        call_endpoint(
+            "update_operators",
+            update_operators,
+            client=wrapper_alice,
+            contract=wrapper_alice,
+        )
+
+        # Send some kit back to the main test account
```
gkaracha, line 565:
```suggestion
        # Send some tez tokens back to the main test account
```

e2e/main.py
```diff
@@ -45,7 +45,18 @@ def assert_kit_balance(checker: ContractInterface, address: str, expected_kit: i
...
+    if expected_kit != kit_balance:
+        raise AssertionError(f"Expected {expected_kit} but got {kit_balance}")
+
+
+def assert_wrapped_tez_balance(wrapper: ContractInterface, address: str, expected: int):
+    fa2_balance_of = {
+        "requests": [{"owner": address, "token_id": 2}],
+        "callback": None,
+    }
+    balance = wrapper.balance_of(**fa2_balance_of).callback_view()[0]["nat_2"]
```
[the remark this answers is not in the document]
dorranh: No worries, that sounds good!

Issue #271, opened 2021-10-12
token ids and src modules


dorranh:
As discussed in #269 and #270, we now have at least three different tokens to keep track of, each with their own token id: kit, wrapped tez, and liquidity tokens. This makes it tough writing entrypoints and end to end tests which interact with either checker or the tez wrapper's FA2 interfaces since we have to be careful that we specify the correct token id. Furthermore, the token id is something that should be easily configurable for collateral tokens since it is likely to be different for the various FA2 tokens which might be used as collateral.

After some discussions with @gkaracha today, we think that it would be cleaner to: 
1. Use a constant value wherever possible for token ids to avoid hard-coding values and 
1. Specify token ids in `checker.yaml`, generating the various token modules from it (`tok.ml`, `kit.ml`, `lqt.ml`). This comes with two additional advantages: The ability to programmatically determine token ids in tests and a more straightforward method for generating Checker's contract metadata than the current method employed in our build scripts. We can also piggy back on this change to make token scaling factors configurable through the same workflow.

gkaracha:
@dorranh I think that #272 addressed this issue. Are there parts that still remain to be implemented? If not, feel free to close it :slightly_smiling_face:

dorranh:
Ah good catch, this is indeed fixed in #272

PR #272, opened 2021-10-13
Generate token modules from common template

dorranh:
This PR addresses #271 by generating `tok.ml`, `kit.ml`, and `lqt.ml` from the same common template. This template uses a union of the methods defined in each of the current token modules since some of the modules (e.g. `kit.ml`) require some additional functions. 

Some remaining TODOs:
- [x] Remove old references to token ids in code in favor of the ones in the generated code (e.g. `kit_token_id`)
- [x] Use the configured token ids (and name/symbol) for generating contract metadata
- [x] Remove old build stages which try to extract token ids from the various token modules
- [x] Rebase on master once #269 is merged
- [x] Find a way to inject tok_token_id in Burrow contract definition :thinking:

### DIFF
--- a/src/kit.mli
+++ b/src/kit.mli
@@ -1,0 +1,39 @@
+type kit
+
+val kit_token_id : Fa2Interface.fa2_token_id
+
+val kit_decimal_digits : Ligo.nat
+val kit_scaling_factor_int : Ligo.int
+val kit_scaling_factor_nat : Ligo.nat
+
+val kit_add : kit -> kit -> kit
+val kit_sub : kit -> kit -> kit
+
+val kit_min : kit -> kit -> kit
+val kit_max : kit -> kit -> kit
+
+val kit_zero : kit
+val kit_one : kit
+
+val kit_of_denomination : Ligo.nat -> kit
+val kit_to_denomination_int : kit -> Ligo.int
+val kit_to_denomination_nat : kit -> Ligo.nat
+
+val kit_of_fraction_ceil : Ligo.int -> Ligo.int -> kit
+val kit_of_fraction_floor : Ligo.int -> Ligo.int -> kit
+val kit_scale : kit -> FixedPoint.fixedpoint -> kit
+
+val leq_kit_kit : kit -> kit -> bool
+val geq_kit_kit : kit -> kit -> bool
+
+val lt_kit_kit : kit -> kit -> bool
+val eq_kit_kit : kit -> kit -> bool
+val gt_kit_kit : kit -> kit -> bool
+
+(* BEGIN_OCAML *)
+val kit_to_ratio : kit -> Common.ratio
+val kit_compare : kit -> kit -> int
+
+val show_kit : kit -> string
+val pp_kit : Format.formatter -> kit -> unit
+(* END_OCAML *)
--- a/src/lqt.mli
+++ b/src/lqt.mli
@@ -1,0 +1,39 @@
+type lqt
+
+val lqt_token_id : Fa2Interface.fa2_token_id
+
+val lqt_decimal_digits : Ligo.nat
+val lqt_scaling_factor_int : Ligo.int
+val lqt_scaling_factor_nat : Ligo.nat
+
+val lqt_add : lqt -> lqt -> lqt
+val lqt_sub : lqt -> lqt -> lqt
+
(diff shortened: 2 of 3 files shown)


Earthfile
```diff
@@ -183,8 +183,14 @@ generate-code:
...
     SAVE ARTIFACT ./src/tok.ml AS LOCAL src/tok.ml
     SAVE ARTIFACT ./src/tok.ml /
+    SAVE ARTIFACT ./src/kit.ml AS LOCAL src/kit.ml
+    SAVE ARTIFACT ./src/kit.ml /
+    SAVE ARTIFACT ./src/lqt.ml AS LOCAL src/lqt.ml
+    SAVE ARTIFACT ./src/lqt.ml /
     SAVE ARTIFACT ./src/constants.ml AS LOCAL src/constants.ml
     SAVE ARTIFACT ./src/constants.ml /
+    SAVE ARTIFACT ./src/burrowOrigination.ml AS LOCAL src/burrowOrigination.ml
+    SAVE ARTIFACT ./src/burrowOrigination.ml /
```
gkaracha, line 193: Ah, yes, sorry, I forgot to change the `Earthfile` when moving the burrow origination into a separate module, thanks for addressing this :+1: I missed it also when I separated the drift derivative calculation (79de464c51c1b8855a6456e546915a6119544954); I have to patch that too :frowning_face:

client/checker_client/cli.py
```diff
@@ -179,20 +179,21 @@ def deploy(config: Config, address=None, port=None, key=None):
...
+def checker(config: Config, checker_dir, oracle, tez_wrapper, ctez, checker_config):
     """
     Deploy checker. Requires addresses for oracle and ctez contracts.
     """
     if not config.oracle_address and not oracle:
         raise ValueError(
             "Oracle address was neither specified in the CLI config nor provided as an argument."
         )
-    if not config.tez_wrapper and not tez_wrapper:
+    if not config.tez_wrapper_address and not tez_wrapper:
```
gkaracha, line 196: :pray:

scripts/generate-ligo.sh
```diff
@@ -48,10 +49,10 @@ tez_wrapper_sources=(
   error
   common
   fixedPoint
-  kit
-  lqt
   fa2Interface
   fa2Ledger
+  kit
+  lqt
```
gkaracha, line 55: Now that I look at it actually, I can't find a reason to have `kit`, `lqt`, or even `fixedPoint` in this list at the moment (i.e., I see no dependency on those for the tez wrapper) :thinking: Not necessarily important for this PR but we should probably remove them.

src/liquidationAuction.ml
```diff
@@ -280,7 +280,7 @@ let start_liquidation_auction_if_possible
     let split_threshold =
       (* split_threshold = max (max_lot_size, FLOOR(queued_amount * min_lot_auction_queue_fraction)) *)
       let { num = num_qf; den = den_qf; } = min_lot_auction_queue_fraction in
-      max_tok
+      tok_max
```
gkaracha, line 283: Ha, inconsistently named? Nice catch! :+1:

gkaracha:
> Thanks for the review @gkaracha. Just pushed an update to the assertion function in the e2e tests.

Thanks! :pray: 

> Regarding the interface files, I think this actually raises a good question. I had removed them since while they ensure that we aren't missing public functions, they also potentially give a false sense of safety in cases where there are private entities in OCaml which are actually public when transpiled to Ligo (due to the lack of module interfaces in Ligo). I wasn't sure if the first property outweighed the downside. What do you think?

Well, you're right about the "false sense of security" issue, but it is ever-present in our OCaml-to-LIGO translation so I wouldn't really worry about that; we have no sense of security whatsoever at this point is what I mean. However, the code generator **_must_** match an API (the rest of the codebase really expects access to `tok_token_id` etc.) so it's important to make that clear in case we need to change the generator later. Indeed, it wasn't encapsulation that I was aiming for, it was lower maintenance troubles :slightly_smiling_face: You're right that one could still define more functions within the module and they would end up in the generated LIGO code.

However, if there is an `.mli` file and we make a breaking change in the generator later, the error will point to the generated file, not the call sites (as is the case if you omit the `.mli` file). I find this to be much cleaner overall.

PR #276, opened 2021-10-18
Call ensure_no_tez_given only once in checkerMain

### LINKED ISSUE #275: Remove duplication around receiving unwanted tez
Currently most of Checker's entrypoints call `Common.ensure_no_tez_given` individually to prevent accidental fund loss. Some of the entrypoints (such as `%deposit_collateral`) don't, of course, since they are meant to receive tez. Once #269 is merged, there will no longer be a need for Checker to deal with tez in any of its entrypoints. It might be a better idea thus to call `Common.ensure_no_tez_given` only once in `checkerMain.ml`. This should help us reduce duplication in both the codebase and the tests. One concern that I have is whether gas costs will go up or down (entrypoint code is serialized, while the contents of checkerMain are not serialized); we'll have to evaluate that.

gkaracha:
Initially this change will break a bunch of tests, but that's OK. What is more important is to evaluate whether gas costs overall go up or down:
* If by checking for unwanted tez in CheckerMain gas costs go down then we can reduce duplication and patch the tests.
* If by sharing the checks go up though, then we should simply make sure that ALL entrypoints check for unwanted tez, add tests for entrypoints that have been until now untested in this respect, and otherwise leave things as they are.

Either way, this PR shall close #275.

### DIFF
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -131,9 +131,6 @@ let[@inline] ensure_valid_leaf_ptr (mem: mem) (leaf_ptr: leaf_ptr) : unit =
   | Some (Leaf _) -> ()
   | _ -> Ligo.failwith error_InvalidLeafPtr
 
-(* TODO: if no tez is allowed to be given anywhere, it might be a better idea
- * to check it once in checkerMain.ml instead of in each entrypoint. *)
-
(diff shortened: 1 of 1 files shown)


dorranh:
> gas costs go down slightly by sharing the check

Awesome, that's great news!

Issue #261, opened 2021-10-01
Generalization of the drift derivative calculation


gkaracha:
(This is Part 2 of the initial generalization issue #171.)

Currently we use a _"bang-bang"_ calculation of the drift derivative ([docs](https://hackmd.io/teMO2x9PRRy1iTBtrSMBvA#Algorithmic-control), [code](https://github.com/tezos-checker/checker/blob/6b0035d3191493e662abdd6a5770c6428ac35126/src/parameters.ml#L126-L207)). However, it would be nice if the curve for the calculation of the drift derivative could be chosen at build time. More specifically, we can offer
* a choice between two curves: the current one and a continuous one, and
* the ability to choose the parameters of the selected curve.

PR #281, opened 2021-10-29
Scaffolding for configurable drift derivative curves

dorranh:
This PR partially implements the configurable drift derivative calculation out lined in https://github.com/tezos-checker/checker/issues/261. It adds a new section to the `checker.yaml` for specifying (known) drift derivative curves and their parameters. This is then used to generate `src/driftDerivative.ml` using a corresponding template, which is where the existing drift derivative module has been moved. Attempts to use `continuous` instead of `bang-bang` in the config file are set to fail at the moment since we haven't yet implemented the continuous drift derivative module.

It also tweaks the existing dataclasses in `checker_builder.config` to make them immutable. This is only really necessary for the new ones added in this PR, but it felt better to add them across the board.


gkaracha (approved):
Slick! I'll go ahead and merge this one right away :ok_hand:

PR #282, opened 2021-11-01
Utilize the received price of ctez in tez

### LINKED ISSUE #239: Deriving the price of kit in tez
When touching the parameters (in `Checker.touch_with_index`) we need the price of kit in tez (units: tez/kit). This we can get by multiplying
1. the price of kit in ctez (units: ctez/kit), as provided by `Cfmm.cfmm_kit_in_ctez_in_prev_block`, and
2. the price of ctez in tez (units: tez/ctez), as provided by the ctez contract.

When we changed the cfmm to use ctez instead of tez and removed the delegation auction we refrained from further complicating matters and assumed that the price of ctez in tez was exactly 1 (i.e., ctez ~ tez). I reckon it's time we switched to a more accurate calculation. I can see three possible ways to deal with (2) above:
1. Assume tez/ctez = 1 (current approach). Simple, since no calls to the ctez contract are required, but probably inaccurate.
2. Assume tez/ctez = 1.05 - 1.06 (that is, the somewhat expected target factor of [ctez](https://github.com/tezos-checker/ctez)). Simple, since no calls to the ctez contract are required. Hopefully rather accurate, but probably too firm.
3. Ask the ctez contract what the current tez/ctez price is. This is clearly the most complicated but at the same time the most accurate option.

gkaracha:
This is the continuation / last part of #280, which should close #239.

### DIFF
--- a/src/checker.ml
+++ b/src/checker.ml
@@ -902,15 +902,14 @@ let entrypoint_receive_price (state, price: checker * Ligo.nat) : (LigoOp.operat
   else
     (([]: LigoOp.operation list), {state with last_index = Some price})
 
-let entrypoint_receive_ctez_marginal_price (state, _price: checker * (Ligo.nat * Ligo.nat)) : (LigoOp.operation list * checker) =
+let entrypoint_receive_ctez_marginal_price (state, price: checker * (Ligo.nat * Ligo.nat)) : (LigoOp.operation list * checker) =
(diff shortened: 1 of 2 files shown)


dorranh (approved):
Thanks for updating this. Looks good to me!

gkaracha:
Great, thanks for having a look!

Issue #279, opened 2021-10-22
Implement an FA2 wrapper for ctez


gkaracha:
Now that we've switched to the tez-wrapper-based design, the biggest remaining obstacle for finishing #213 is dealing with the structural differences between TEZ and FA2 that we've found no way to share between the two build/deployment paths. As far as I can tell, this boils down to two specific issues, interrelating with each other:

1. Price calculation in `touch`:
    - `collateral = TEZ`: KIT/CHF = TEZ/CHF (from index) * CTEZ/TEZ (from ctez) * KIT/CTEZ (from cfmm)
    - `collateral = FA2`: KIT/CHF = FA2/CHF (from index) * KIT/FA2 (from cfmm)

    this issue is discussed here: https://github.com/tezos-checker/checker/issues/213#issuecomment-923758356

2. The token contained in the CFMM compared to the token used as collateral:
    - `collateral = TEZ`: `cfmm.ml` contains ctez **_specifically_**, also distinct from the collateral which is tez_token **_specifically_**
    - `collateral = FA2`: `cfmm.ml` contains the same token as is used for collateral, whichever token that is.

The [comment on #213](https://github.com/tezos-checker/checker/issues/213#issuecomment-923758356) is primarily focused on the first issue, but, while trying to make progress on the the second issue (on [this branch](https://github.com/tezos-checker/checker/compare/gkaracha/cfmm-token)), I came across another structural difference:

3. Token standard used for fund transfer:
    - `collateral = TEZ`: `cfmm.ml` contains ctez and it uses ctez's [FA1.2 API][fa12]
    - `collateral = FA2`: `cfmm.ml` contains the same FA2 that is used as collateral so **_I expect_** it'll use the token's [FA2 API][fa12]

FA1.2 and FA2 are structurally different, so the branching as stated above is quite pervasive (i.e., we'll need conditional compilation or something equivalent to get generalization to work). Ideally we won't have to go that way. However, as far as I can tell, ctez does not offer an FA2 interface. To achieve uniformity in the `%transfer` calls we'd have to always use the FA1.2 standard for the cfmm, but then we require the token used to respect both the FA2 and the FA1.2 standard.

This looks like a blocker for #213.

[fa2]: https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-12/tzip-12.md
[fa12]: https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-7/tzip-7.md

gkaracha:
[...]

PR #285, opened 2021-11-04
Make cfmm.ml (mostly) agnostic to the kind of FA2 (currently FA1.2) used

gkaracha:
This is a somewhat bikesheddy PR, but it's necessary preparatory work before we are able to switch between the collateral being TEZ or FA2. It implements most of the changes needed to address the second issue outlined in #279.

Commit-log:
- [X] Create `ctok.ml` (i.e., "CFMM token")
- [X] Replace uses of `ctez` with uses of `ctok` in the CFMM (remove `ctez.ml` altogether).
- [X] Reconcile `genericToken.ml.jinja` and `ctok.ml`.
- [X] Generate ctok.ml from the yaml configuration.
- [X] Update the docs (`functional_spec.rst` in particular).

Sidenote: The first wave of real API changes will probably come when we implement #279.


gkaracha:
@dorranh due to the renamings this is a pretty noisy PR, but I tried to keep each commit self-contained and CI-passing. Might be easier reviewing this PR per commit :sweat_smile:

dorranh (approved):
It'll be great to have all of this already refactored as we go to actually change the cfmm logic. I had a look, and it looks good to me! We might end up needing to tweak the schema of the yaml file a bit once we go to actually implement the tez / fa2 build switch, but I think we can go ahead and land this PR as-is.

gkaracha:
Great, thanks! :pray:

Yes, after merging this I think we have to first land #286, and then change Checker to actually go through the ctez wrapper, before even adding the switch. I am thinking that `price.ml` will also have to change at the same time as when we add the switch, so we probably have to turn it into a jinja template as well before the switch :thinking:

PR #286, opened 2021-11-05
Add FA2 contract which wraps ctez

dorranh:
Addresses #279 by implementing an FA2 contract which issues wrapped ctez tokens (`wctez`). This can then be used directly in Checker's CFMM to enable uniform internal operations for both the tez collateral case (which will use `wctez`) and general FA2 collateral.

### DIFF
--- a/src/wctez.ml
+++ b/src/wctez.ml
@@ -1,0 +1,205 @@
+(* FA2 wrapper for ctez to allow it to integrate with Checker *)
+open Fa2Interface
+open Fa2Ledger
+open Fa12Interface
+open Common
+open Error
+
+(*****************************************************************************)
+(**                          {1 WRAPPER TYPES}                               *)
+(*****************************************************************************)
+
+type wctez_state =
+  { fa2_state : fa2_state;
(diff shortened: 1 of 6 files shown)


src/wctez.ml
```diff
@@ -0,0 +1,205 @@
...
+  }
+
+(** Token id for wrapped ctez tokens. *)
+let[@inline] wctez_token_id : fa2_token_id = Ligo.nat_from_literal "0n"
+
+(*
+(** Number of decimal digits for wctez tokens, identical to that for ctez. *)
+(* NOTE: Currently unused. *)
+let[@inline] wctez_token_decimal_digits = Ligo.nat_from_literal "6n"
+*)
```
gkaracha, line 24: Ah yes, same as with `tezWrapper.ml`. We have to discuss at some point how to make the metadata of our two wrappers available on chain. Currently only Checker advertises the metadata of `kit` and `lqt` but the two wrappers should probably do the same as well.

gkaracha (approved):
Flawless :ok_hand: The gas costs are very low as well; feel free to merge this one as-is!

PR #292, opened 2021-11-11
Turn the price calculation into a jinja template

gkaracha:
Other changes
* To avoid some cyclic dependencies I moved some utility functions from `checker.ml` to `checkerTypes.ml`, since they pertain to the `external_contracts` type.
* Renamed `kit_in_tez` to `kit_in_tok` in places where it was misspelled; there's no tez anywhere in Checker at the moment, `tok` captures the collateral type.
* Added `src/price.mli`
* BugFix: replaced `tez_scaling_factor_int` by `tok_scaling_factor_int` and removed `tez_scaling_factor_int` altogether; these places would erroneously use a scaling factor of 1_000_000 even if the FA2 token used as collateral has a different scaling factor. While at it, I also removed `fraction_to_tez_floor` and `tez_scaling_factor_nat` that were no longer needed either.
* Also added `scripts/builder/checker_builder/templates/fa2Price.ml.jinja` already and checked that we can use it successfully to calculate the price (e2e tests should pass, I mean).


dorranh (approved):
LGTM! No comments other than that it'll be good to have pruned some of the unused and inaccurately named code.

gkaracha:
Thanks!

Issue #291, opened 2021-11-09
Single source of truth for token metadata


gkaracha:
In principle, all configuration should come from `checker.yaml`. Currently we have a mix of hard-wired and configurable information, which can easily get out of sync. Taken from https://github.com/tezos-checker/checker/pull/290#issuecomment-964113677:

> Also, while developing this I came across some issues due to the way we currently deal with a couple token ids:
> * `tokens.collateral.token_id` (in `checker.yaml`) and `TezWrapper.tez_token_id` must be the same when `collateral = TEZ`.
> * `tokens.cfmm_token.token_id` (in `checker.yaml`) and `Wctez.wctez_token_id` must be the same when `collateral = TEZ`.
> Currently `TezWrapper.tez_token_id` and `Wctez.wctez_token_id` are set manually which is a problem (when they get out of sync Checker gets into an inconsistent state and the e2e tests fail).

Let's have **_all_** token information come from `checker.yaml`, even for those tokens that are conditionally generated (both `wctez` and `tezWrapper` are needed when `collateral = TEZ`, which is the only case at the moment).

gkaracha:
`mockFA2` (#288) might also fall into this category, even if it is to be used only for testing purposes.

PR #293, opened 2021-11-11
Refactor tokens section in checker.yaml

dorranh:
This PR takes a first step towards addressing #291 by refactoring the `tokens` section of `checker.yaml` and corresponding build script logic, adding fields for `wtez`, `wctez`, `mock_fa2`. It also reorganizes the `tokens` schema a bit, distinguishing between `issued` and `in_use` tokens since tokens in the former category need to allow for configurable FA2 metadata.

I also went ahead and added a field for the `collateral_type` switch and made the build script throw an error if it is ever not set to `tez` since we haven't implemented the FA2 build logic yet.

After this PR I should be able to implement the generation of a new `tokenMetadata.ml` module we discussed which can be used within `tezWrapper.ml` and `wctez.ml` (and other generated token modules).


scripts/builder/checker_builder/config.py
```diff
@@ -114,10 +143,37 @@ class Constants:
...
+        # Validation logic
+        wtez = self.tokens.issued.wtez
+        wctez = self.tokens.issued.wctez
+        collateral = self.tokens.in_use.collateral
+        cfmm_token = self.tokens.in_use.cfmm_token
+
+        if self.collateral_type == CollateralType.TEZ:
+            if not (collateral.token_id == wtez.token_id) and (
+                collateral.decimal_digits == wtez.decimal_digits
+            ):
```
gkaracha, line 161: Hmmm, shouldn't the not scope over the whole condition? I mean like this:

```suggestion
            if not ((collateral.token_id == wtez.token_id) and (
                collateral.decimal_digits == wtez.decimal_digits
            )):
```

scripts/builder/checker_builder/config.py
```diff
@@ -114,10 +143,37 @@ class Constants:
...
+                raise ValueError(
+                    "collateral config must be identical to wtez config when collateral_type=tez"
+                )
+            if not (cfmm_token.token_id == wctez.token_id) and (
+                cfmm_token.decimal_digits == wctez.decimal_digits
+            ):
```
gkaracha, line 167: Same here :thinking:

gkaracha (commented):
Looks good to me overall, except for the negation in the if-then-else conditions (see below); lemme know if I got that wrong :sweat_smile: No idea what's up with CI but let's not merge before that is resolved.

scripts/builder/checker_builder/config.py
dorranh: Ah yes it absolutely should. Good catch!

checker.yaml
```diff
@@ -1,50 +1,83 @@
...
+      token_id: 2
+      # Specifies the number of decimal digits used by the token
+      decimal_digits: 6
+    # CFMM token (creates ctok.ml).
+    # Note: Must be configured with same values as wctez when collateral_type=tez
+    cfmm_token:
+      # The token's FA2 id
+      token_id: 42
+      # Specifies the number of decimal digits used by the token
+      decimal_digits: 6
```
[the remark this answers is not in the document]
dorranh: I think that for the current configuration with `collateral_type: tez` this has to match `wctez`. Definitely agree for once we have implement support for `fa2` collateral though.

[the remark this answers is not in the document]
gkaracha: Ah, sorry, I didn't mean this one here, no idea how I ended up leaving this comment :facepalm: Scrap it.

gkaracha (approved):
LGTM! :+1:

PR #296, opened 2021-11-16
Make checker.yaml the single source of truth for token metadata

### LINKED ISSUE #291: Single source of truth for token metadata

gkaracha:
Closes #291.

Turning this into a PR early, to see if there are any unexpected gas cost changes due to the new module dependency graph.

Apart from the main change (i.e., moving `token_id`s and `token_decimal_digits` into a dynamically generated file), this PR also implements some small renamings in the relevant files, for uniformity. (A bug was also found in mockFA2 module, in the process.)

~~**REMAINING TODO**: Update `e2e/main.py`; this is the only remaining file that hardcodes some token_ids.~~ Addressed.


scripts/builder/checker_builder/cli.py
```diff
@@ -58,6 +61,10 @@ def generate():
     # Select the price calculation module at runtime based on config
     GENERATE_SRCS[PRICE_SRC] = PRICE_TEMPLATES[checker_config.collateral_type]
 
+    # Generate the token metadata module at runtime based on config
+
+    GENERATE_SRCS[TOKEN_METADATA_SRC] = TOKEN_METADATA_TEMPLATE
```
dorranh, line 66: Since the template isn't selected at runtime, I think that we can actually move this into the static `GENERATE_SRCS` on  line 8. The dynamic part which comes from the config will already get passed when the template is rendered below.

dorranh (commented):
Thanks, @gkaracha! Just one small comment regarding the organization of `checker_client.checker`. I'm happy merging once that is addressed.

scripts/builder/checker_builder/cli.py
gkaracha: Ah, right, good point! Done in bd8a9fb :+1:

gkaracha:
Great, thanks! I addressed your comment; I shall merge once CI is green :slightly_smiling_face:

Issue #295, opened 2021-11-15
Include metadata in the storage of our helper contracts


gkaracha:
This pertains to the tez wrapper (`wtez`) and the ctez wrapper (`wctez`). The mock FA2 contract does not need them, strictly-speaking, but when we add the metadata for the other two we can probably do so for that one as well.

Related: #291.

PR #298, opened 2021-11-17
Add metadata fields to wtez, wctez, and mockFA2

gkaracha:
This constitutes the static/structural part of #295. Next we need to populate the metadata at deployment with bytes generated from `checker.yaml`.

### DIFF
--- a/src/mockFA2.ml
+++ b/src/mockFA2.ml
@@ -6,7 +6,7 @@ open Fa2Ledger
 open Common
 open TokenMetadata
 
-type params =
+type mock_fa2_params =
   (* FA2 entrypoints *)
   | Balance_of of fa2_balance_of_param
   | Transfer of fa2_transfer list
@@ -14,6 +14,11 @@ type params =
   (* Contract-specific entrypoints *)
   | Mint of Ligo.nat
   | Redeem of Ligo.nat
+
+type mock_fa2_state =
+  { fa2_state : fa2_state;
+    metadata: (string, Ligo.bytes) Ligo.big_map;
+  }
 
 let[@inline] fa2_get_balance (st, owner, token_id: fa2_state * Ligo.address * fa2_token_id): Ligo.nat =
   let ledger = st.ledger in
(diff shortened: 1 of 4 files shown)


src/mockFA2.ml
```diff
@@ -66,8 +70,10 @@ let[@inline] fa2_run_transfer (state, xs: fa2_state * fa2_transfer list) : fa2_s
       xs in
   (state, ([]: LigoOp.operation list))
 
-let[@inline] transfer (state: fa2_state) (xs: fa2_transfer list) : LigoOp.operation list * fa2_state =
-  let state, ops = fa2_run_transfer (state, xs) in
+let[@inline] transfer (state: mock_fa2_state) (xs: fa2_transfer list) : LigoOp.operation list * mock_fa2_state =
+  let { fa2_state = fa2_state; metadata = metadata; } = state in (* deconstruct *)
+  let fa2_state, ops = fa2_run_transfer (fa2_state, xs) in
+  let state = { fa2_state = fa2_state; metadata = metadata; } in (* reconstruct *)
```
dorranh, line 76: Minor point, but it might be good to keep the syntax uniform for handling the reconstruction of the final state (we use this approach here and `{st with fa2_state = fa2_state}` in `wctez.ml`). Not sure which is better, but perhaps the latter reduces the likelihood of accidentally messing with the metadata.

src/wtez.ml
```diff
@@ -138,16 +139,16 @@ let[@inline] fa2_run_transfer (st, xs: wtez_state * fa2_transfer list) : wtez_st
...
     Ligo.List.fold_left
       (fun (((st, ops), tx): (wtez_state * LigoOp.operation list) * fa2_transfer) ->
-         let { fa2_state = fa2_state; vaults = vaults; } = st in (* deconstruct *)
+         let { fa2_state = fa2_state; vaults = vaults; metadata = metadata; } = st in (* deconstruct *)
          let { from_ = from_; txs = txs; } = tx in
 
          (* Origination of the from_ vault, if needed *)
          let from_vault_found, ops, vaults, from_vault_address = find_vault_address_append vaults from_ ops in
-         let st = { fa2_state = fa2_state; vaults = vaults; } in (* reconstruct *)
+         let st = { fa2_state = fa2_state; vaults = vaults; metadata = metadata; } in (* reconstruct *)
```
dorranh, line 147: Same question for this module

dorranh (approved):
Nice just one small comment, but otherwise LGTM.

src/mockFA2.ml
gkaracha: Ah, nice point, yes. In general explicit deconstruction/reconstruction I think is more efficient (gas-wise).

The field update reduces the likelihood of accidentally messing with the metadata, true, but it's pretty hard to create bytes in our current setup so I wouldn't worry about that very much. 

On the other hand the field update will type-check without errors/warnings when we update the state type which can be dangerous (especially in the case of the tez wrapper, where sometimes we need to update the vaults and the fa2_state, but other times we only need to update the fa2_state). Given this, I'd uniformly switch to explicit deconstruction/reconstruction.

dorranh: Okay great, sounds good to me!

gkaracha: Done in c5f239a!

I noticed that we still use field updates in `fa2_run_update_operators` (for all FA2 modules) but I am not gonna risk touching that part unnecessarily, since it does not relate to the changes in this PR (and, more importantly, I'd hate to introduce bugs now just to refactor this :sweat_smile:)

PR #299, opened 2021-11-17
Use same configuration file during build and deployment

dorranh:
Note: this is currently still a bit of an experiment.

Updates `checker_client` and `checker_builder` to store a copy of the configuration file used during code generation and load that one specifically during deployment. 

As a part of this I tinkered with how we currently construct paths in our various Python code, which at the moment includes relative paths throughout it. In an attempt to make things more uniform (and less annoying to refactor), I've moved the relative path logic into a single class `CheckerRepo` and made most of the methods which previously operated on individual paths take that as an input instead. This makes the code more tightly coupled to our directory structure (i.e. the functions will only run within the predefined directory structure of our repo), but this is really only making explicit what was previously assumed throughout the Python modules. IMO you wouldn't really want to use the CLI to interact with items outside of the context of this repo anyways since we have a fairly complicated build and deployment process.


gkaracha (approved):
Thanks @dorranh! This one looks good to me as-is, I'll just go ahead and merge :slightly_smiling_face:

PR #301, opened 2021-11-23
Ensure that `earthly +test` and `earthly +test-e2e` work with both collateral configurations

gkaracha:
* Create default configurations `checker-tez.yaml` and `checker-fa2.yaml`. At build/deployment time we can just copy the one we wish to use into `checker.yaml`.
* Fix references to the wrong token id in `WTezTest`.
* Patch the end-to-end tests so that they can function under both configurations for `collateral_type`.
* Comment out a couple of tight unit tests that emit different operations under different configurations. (Ideally we'd re-enable these tests and have two separate testsuites, one per collateral type).

While updating `client/checker_client/cli.py` I also noticed that the commands for deployment we currently have in the README.md don't work on my machine due to a failure to deploy ctez's CFMM. This needs further investigation.


gkaracha:
I'll go ahead and merge this one. It'd be nice to have CI use the two default configurations (as I crudely tried on 312ebe0737e692a039b8daa43eeea541045c8f37) but CI hours effectively double and I'm not sure whether that's OK. At least after merging this PR we can do that locally when we need to :slightly_smiling_face:
```bash
cp checker-fa2.yaml checker.yaml && earthly +build-ocaml
```

PR #302, opened 2021-11-24
Update checker-client Docker image and deploy instructions

dorranh:
Updated the checker-client Docker image build to:
1. Fix a broken PATH variable which prevented the `checker` command from being found
2. Updated the working directory to allow deploy commands to function more smoothly for bundled contracts
3. Added some missing system dependencies

Also added a helper script for using this image to deploy Checker and its contracts and updated the corresponding sections in our README.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -402,7 +402,7 @@ cli:
 
     ENV DEBIAN_FRONTEND=noninteractive
     RUN apt update && apt install -y \
-          pkg-config autoconf libtool libev4 \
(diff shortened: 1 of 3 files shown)


dorranh:
Converting to a draft since while the PATH updated here looks correct, the working directory looks off (I think the default working directory needs to be `/root` instead of `/root/client` for the defaults for the `checker deploy` commands to work smoothly.

dorranh:
Going ahead and merging this one following offline discussions with @gkaracha.

PR #306, opened 2021-12-02
Add TZIP16 metadata for wtez, wctez, and the mock FA2 contract

### LINKED ISSUE #295: Include metadata in the storage of our helper contracts

gkaracha:
Closes #295


gkaracha:
@dorranh I'm turning this into a draft PR in case you wanna have a quick look (I'll ping you once it's done anyway so no hurry). There's one last thing missing that is a little annoying: [`view_total_supply`](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-12/tzip-12.md#off-chain-views) is meant to _"return [the] total number of tokens for the given token-id if known or fail if not"_. To be able to implement that we have to keep track of these values in the storage of each contract. Pity, since we don't really have any other use for the total amount of token in circulation, but it could come in handy for testing later (like we do in checker using `fa2_get_total_kit_balance` and `fa2_get_total_lqt_balance`).

gkaracha:
Hey @dorranh this should be ready for review now, assuming CI will be green :slightly_smiling_face:

client/checker_client/checker.py
```diff
@@ -294,14 +336,28 @@ def deploy_wtez(
...
+    config = load_input_config()
     src = repo.wtez_contract
+
+    with repo.wtez_metadata.open() as f:
+        views = json.load(f)["views"]
+
+    # Compute the TZIP-16 metadata
+    token_metadata_view = wtez_token_metadata_view_from_config(config=config)
+    metadata = tzip16_metadata_from_views(views + [token_metadata_view])
+    metadata_ser = json.dumps(metadata).encode("utf-8")
```
dorranh, line 348: Out of curiosity, is storing the metadata as a JSON string part of the FA2 spec?

src/mockFA2.ml
```diff
@@ -117,21 +118,26 @@ let[@inline] fa2_run_update_operators
...
   (([]: LigoOp.operation list), state)
 
 let[@inline] redeem (state: mock_fa2_state) (amnt: Ligo.nat) : LigoOp.operation list * mock_fa2_state =
-  let { fa2_state = fa2_state; metadata = metadata; } = state in (* deconstruct *)
+  let { fa2_state = fa2_state; total_token = total_token; metadata = metadata; } = state in (* deconstruct *)
   let fa2_state = ledger_withdraw (fa2_state, mock_fa2_token_id, !Ligo.Tezos.sender, amnt) in
-  let state = { fa2_state = fa2_state; metadata = metadata; } in (* reconstruct *)
+  let total_token =
+    match Ligo.is_nat (Ligo.sub_nat_nat total_token amnt) with
+    | None -> (failwith "FA2_INSUFFICIENT_BALANCE" : Ligo.nat)
```
dorranh, line 138: I'm wondering if it's okay to throw this error here or if we need something more specific. I couldn't find any relevant details in TZIP-12 other than that this should be thrown when transfers exceed the senders balance in the ledger.

src/wctez.ml
```diff
@@ -183,7 +184,12 @@ let[@inline] redeem (state: wctez_state) (amnt: Ligo.nat) : LigoOp.operation lis
   in
   (* Remove the specified amount of tokens for the caller *)
   let fa2_state = ledger_withdraw_wctez_token (fa2_state, !Ligo.Tezos.sender, amnt) in
-  let state = { fa2_state = fa2_state; ctez_fa12_address = ctez_fa12_address; metadata = metadata; } in (* reconstruct *)
+  (* Remove the specified amount of token from circulation *)
+  let total_token =
+    match Ligo.is_nat (Ligo.sub_nat_nat total_token amnt) with
+    | None -> (failwith "FA2_INSUFFICIENT_BALANCE" : Ligo.nat)
```
dorranh, line 190: Same question as above

src/wtez.ml
```diff
@@ -5,8 +5,6 @@ open VaultTypes
 open Error
 open TokenMetadata
 
-(* TODO: Expose FA2 metadata for tez_tokens. *)
```
dorranh, line 8: It's nice to be able to remove some of these TODOs! :slightly_smiling_face:

src/wtez.ml
```diff
@@ -234,66 +233,71 @@ let[@inline] fa2_run_update_operators
...
     ([origination; op], state)
 
 let[@inline] withdraw (state: wtez_state) (amnt: Ligo.tez) : LigoOp.operation list * wtez_state =
-  let { fa2_state = fa2_state; vaults = vaults; metadata = metadata; } = state in (* deconstruct *)
+  let { fa2_state = fa2_state; total_token = total_token; vaults = vaults; metadata = metadata; } = state in (* deconstruct *)
   let _ = ensure_no_tez_given () in
   let fa2_state = ledger_withdraw_tez_token (fa2_state, !Ligo.Tezos.sender, amnt) in
+  let total_token =
+    match Ligo.is_nat (Ligo.sub_nat_nat total_token (tez_to_mutez_nat !Ligo.Tezos.amount)) with
+    | None -> (failwith "FA2_INSUFFICIENT_BALANCE" : Ligo.nat)
```
dorranh, line 276: Same question as above

dorranh (commented):
Nice! It'll be great to have these views available. Just a couple of small questions but otherwise looks good to me. I tried to validate the new regex(s) in `compile-ligo.rb` but am honestly having trouble parsing them in my brain :sweat_smile:

client/checker_client/checker.py
gkaracha: Indeed it is! See the ["Contract Storage" section of TZIP16](https://gitlab.com/tezos/tzip/-/blob/master/proposals/tzip-16/tzip-16.md#contract-storage):

> At least one value must be present:
> * the one for the empty string key (`""`).
> * the value must be a URI as specified in the following section which points to a JSON document as specified further below.
> [..]

client/checker_client/checker.py
```diff
@@ -153,6 +170,31 @@ def compile_view_fa2_token_metadata(tokens: List[TokenMetadata]):
     }
 
 
+# FIXME: TYPE?
```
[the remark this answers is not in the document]
gkaracha: Ah, right. I'll remove the comment then, it's pretty self-evident that it's a dictionary I guess.

src/mockFA2.ml
gkaracha: Right. So, since this is in entrypoint code (not in metadata), TZIP-12 is oblivious to the specific error we throw (and, indeed, TZIP-16 does not specify any error we MUST emit). TZIP-12 does though. Currently I i
mplemented all additions/subtractions in this order:
```
account[user] -= amnt
total_token   -= amnt
```
and not
```
total_token   -= amnt
account[user] -= amnt
```
But I didn't want the code to be sensitive to the order of the two subtractions, so I made them both fail with the correct error mnemonic (i.e., so that we comply with TZIP12 independently of the order of these subtractions). Alternative we could make the `total_token` reduction fail with an internal "impossible-happened" error, but then we'd have to always perform that subtraction **after** the subtraction from the specific account.

dorranh: Gotcha, thanks for the additional background. In that case I'm also happy leaving it as-is. I don't think there is really a huge benefit to throwing a different error code here other than to indicate that the error is an internal one.

dorranh:
@gkaracha, I went ahead and tested out deploying to `granadanet` using this PR and it seems to be working without issue. I think we can go ahead and merge this.

gkaracha:
Thanks for the patches and the e2e tests @dorranh! :pray: I also added a few unit tests on the OCaml side (see 408476e51351309e8db519a76067b96ff1b68a77, 52666759d5a46f0777995668b9b657ac014fac86, and 4042ab42ace526f833d840b3d3ec2402e27964bd) though there are many more we should add. I don't want this PR to linger though so once CI is green I'll go ahead and merge it. We can add more unit and property-based tests incrementally.

dorranh:
No problem, and that sounds good!

PR #307, opened 2021-12-03
Cleaner sealing

gkaracha:
Tiny PR I've been meaning to open for a little while. When sealing the contract we can pass the external addresses as a record instead of an uninformative tuple of addresses `address * address * address * address`. This makes is harder to pass the addresses in the wrong order (which has happened to me in the past during development).

### DIFF
--- a/client/checker_client/checker.py
+++ b/client/checker_client/checker.py
@@ -419,12 +419,17 @@ def deploy_checker(
             print("  deployed: chunk {}.".format(chunk_no))
 
     print("Sealing.")
+
+    external_contracts = {
+        "oracle": oracle,
+        "collateral_fa2": collateral_fa2,
+        "ctok_fa2": cfmm_token_fa2,
+        "ctez_cfmm": ctez_cfmm,
+    }
+
     inject(
         tz,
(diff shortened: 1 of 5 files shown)


client/checker_client/checker.py
```diff
@@ -419,12 +419,17 @@ def deploy_checker(
             print("  deployed: chunk {}.".format(chunk_no))
 
     print("Sealing.")
+
+    external_contracts = {
+        "oracle": oracle,
```
dorranh, line 424: I'm actually surprised that we can pass `ContractInterface` types here instead of the string with the address, but I guess pytezos must convert it internally. I'm happy leaving it since that is what we were doing before anyways.

dorranh (approved):
Nice, this is indeed much cleaner. LGTM!

client/checker_client/checker.py
gkaracha: Right?! I could add the `.context.address` parts if you think important; otherwise feel free to merge :slightly_smiling_face:

dorranh: Ya no worries, it's what we were doing before anyways.

PR #309, opened 2021-12-08
Support for non-root user in dev container

dorranh:
Updates the dev container to use a non-root user. This avoid situations where artifacts created in a volume mount require a `sudo chown` command or similar on the host since they were created using the root user in the container.

In order to accomplish this, I had to expand the entrypoint script a bit to enable re-mapping of the user and docker group within the container at runtime so that they match the respective ids on the host. This requires setting two extra environment variables when launching the contianer, and I updated the README with the new command.

To test locally you will first need to run:
   `earthly +dev-container`

Then modify the command in the README to point to the new local image you just created: `checker/dev:latest`

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -92,6 +92,8 @@ docs:
 spec:
     FROM +builder
     RUN pip install sphinx-rtd-theme
+    # Add local install path for sphinx since we are running as non-root user
+    ENV PATH=/home/checker/.local/bin:$PATH
     COPY docs docs
     RUN make -C docs/spec html
     SAVE ARTIFACT docs/spec/_build/html AS LOCAL docs/spec/_build/html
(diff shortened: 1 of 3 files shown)


gkaracha:
Hey, sorry it took me a little while to get to this. Just rebased it since `Earthfile` had changed a little in the meantime, I'll try it out now and get back to you.

gkaracha (approved):
LGTM! I tested most commands in the README (all but the deployment to a testnet) and all of them work as expected. Managed to delete the repo without issues as well. This should be good to go :slightly_smiling_face:

gkaracha:
PS. I think that the message bot is going crazy :sweat_smile:

dorranh:
Nice, I'm glad to hear it! I'll go ahead and merge.

> PS. I think that the message bot is going crazy

Ya, I think it might need to take some time off for the holidays as well...

Issue #297, opened 2021-11-17
Update protocol to Hangzhou


dorranh:
Once Hangzhou is released on mainnet (I think this should happen within the next week or so), we should upgrade the protocol in our project (used for sandbox, tests, etc.). This will likely involve:

- [x] Updating the version of flextesa we are building in our Earthfile (`+flextesa` target). It looks like flextesa has simplified their build process, so we can also likely simplify our fairly manual build steps at the same time, relying on their `Makefile` instead.
- [x] Updating the protocol in `checker_client.checker.start_local_sandbox`
- [ ] Updating the default protocol in `scripts/compile-ligo.rb`. Alternatively, we can also remove this from the script since we are not measuring the contract size on CI anymore (instead measuring the serialized entrypoints themselves).

dorranh:
Closed in #315 (not that I think that we might want to just remove the `tezos-client` calls in `compile-ligo` since we no longer really use that functionality.

gkaracha:
> not that I think that we might want to just remove the `tezos-client` calls in `compile-ligo` since we no longer really use that functionality.

Ah, yes, this has come up before. I'd be OK with either (a) removing the calls to `tezos-client`, or (b) restoring the `tezos-client` dependency so that we actually get to see again the size the contracts we generate.

Issue #278, opened 2021-10-22
Potential security vulnerabilities in the python dependencies


gkaracha:
Since #272 was merged I've been seeing the following Dependabot warning:
```
We found potential security vulnerabilities in your dependencies.
```
[The warning](https://github.com/tezos-checker/checker/security/dependabot) refers to two python dependencies in our `poetry.lock`:
- [x] [fastecdsa](https://github.com/tezos-checker/checker/security/dependabot/poetry.lock/fastecdsa/open)
- [ ] [pywin32](https://github.com/tezos-checker/checker/security/dependabot/poetry.lock/pywin32/open)

We should probably fix that.

dorranh:
The first point is actually a bit blocked by pytezos which pins an older version of fastecdsa. We can hack around that need be, but the cleaner solution is to upgrade it upstream. I created an issue for this here: https://github.com/baking-bad/pytezos/issues/261

dorranh:
The `pywin32` one is an issue with both our version of `docker-py` and that used by `pytezos`. However, we can't bump our `docker-py` version to the latest release (`5.0.3`) until `pytezos` upgrades since `pytezos` places an upper bound of `<5.0.0`.

dorranh:
Note: the `pywin32` vulnerability is blocked on https://github.com/docker/docker-py/pull/2925.

dorranh:
After reviewing the pywin32 vulnerability a bit more, it seems that this should only affect people running checker's python applications directly on Windows (i.e. on the Windows host itself and not from within a Docker container running on Windows). Since very few of the tools in our stack are supported on Windows, I'm going to go ahead and dismiss that alert. Please feel free to re-open though.

gkaracha:
Sounds good to me! :+1:

PR #310, opened 2021-12-09
Upgrade python packages

dorranh:
Partially addresses #278 by upgrading our `docker` and `pytezos` packages to newer versions (along with a couple of other dependencies which needed to be upgraded in order for everything to resolve). As noted on #278, we'll need to upgrade docker one more time if/when they land our fix for the pywin32 version issue.

This also removes all of the weird `nat_*` references in our end to end tests since name annotations for contract callback views are fixed in newer versions of `pytezos` and is a pre-requisite for the protocol upgrade task (#297).

### DIFF
--- a/e2e/main.py
+++ b/e2e/main.py
@@ -52,7 +52,7 @@ def assert_fa2_token_balance(
         "requests": [{"owner": address, "token_id": token_id}],
         "callback": None,
     }
-    balance = wrapper.balance_of(**fa2_balance_of).callback_view()[0]["nat_2"]
+    balance = wrapper.balance_of(**fa2_balance_of).callback_view()[0]["balance"]
     if expected != balance:
         raise AssertionError(f"Expected {expected} but got {balance}")
 
@@ -543,14 +543,7 @@ class E2ETest(SandboxedTestCase):
             "callback": None,
         }
         print(f"Calling balance_of as an off-chain view with {fa2_balance_of}")
-        balance = checker.balance_of(**fa2_balance_of).callback_view()
-        # Check that this balance agrees with the kit balance view
-        assert_fa2_token_balance(
-            checker,
-            checker_alice.key.public_key_hash(),
-            kit_token_id,
-            balance[0]["nat_2"],
-        )
+        checker.balance_of(**fa2_balance_of).callback_view()
 
         # ===============================================================================
         # CFMM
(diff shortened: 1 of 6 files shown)


e2e/main.py
```diff
@@ -52,7 +52,7 @@ def assert_fa2_token_balance(
         "requests": [{"owner": address, "token_id": token_id}],
         "callback": None,
     }
-    balance = wrapper.balance_of(**fa2_balance_of).callback_view()[0]["nat_2"]
+    balance = wrapper.balance_of(**fa2_balance_of).callback_view()[0]["balance"]
```
gkaracha, line 55: :tada:

e2e/main.py
```diff
@@ -543,14 +548,7 @@ def get_collateral_tokens_and_make_checker_an_operator(amnt):
...
         print(f"Calling balance_of as an off-chain view with {fa2_balance_of}")
-        balance = checker.balance_of(**fa2_balance_of).callback_view()
-        # Check that this balance agrees with the kit balance view
-        assert_fa2_token_balance(
-            checker,
-            account_alice,
-            kit_token_id,
-            balance[0]["nat_2"],
-        )
+        checker.balance_of(**fa2_balance_of).callback_view()
```
gkaracha, line 551: Shouldn't we also check the result of `balance_of` here?

e2e/main.py
```diff
@@ -1345,15 +1348,18 @@ def get_collateral_tokens_and_make_checker_an_operator(checker, amnt):
...
-
-        auction_id, minimum_bid = auction_details["contents"], auction_details["nat_3"]
-        # FIXME: The return value is supposed to be annotated as "auction_id" and "minimum_bid", I
-        # do not know why we get these names. I think there is an underlying pytezos bug
-        # that we should reproduce and create a bug upstream.
-
+        auction_id, minimum_bid = (
+            auction_details["auction_id"],
+            auction_details["minimum_bid"],
+        )
```
gkaracha, line 1362: :tada:

gkaracha (approved):
Finally, no more `nat_2` and `nat_3`! :tada: LGTM overall, I only think that it accidentally removes a check we used to have so best if that comes back before we merge; see comment below.

e2e/main.py
dorranh: Ah, so it turns out that we actually call `balance_of` in the implementation of `assert_fa2_token_balance`, so this check was actually just comparing the same value :upside_down_face:

gkaracha: Haha, nice :stuck_out_tongue:

Still though, since we call
```python
        call_checker_endpoint("transfer", [fa2_transfer])
        assert_fa2_token_balance(checker, account_alice, kit_token_id, 10)
```
a few lines above, this whole bit
```python
        # `balance_of` requires a contract callback when executing on-chain. To make tests
        # more light-weight and avoid needing an additional mock contract, we call it as a view.
        # This comes with a downside, however, since using a view means that we don't get an
        # estimate of the gas cost.
        fa2_balance_of = {
            "requests": [{"owner": account_alice, "token_id": kit_token_id}],
            "callback": None,
        }
        print(f"Calling balance_of as an off-chain view with {fa2_balance_of}")
        checker.balance_of(**fa2_balance_of).callback_view()
```
could go, no? I mean, I don't see any benefit in calling `checker.balance_of(**fa2_balance_of).callback_view()` here

dorranh: Very true. IIRC this ended up this way since we had a different implementation of `assert_fa2_token_balance` originally. I guess the only argument for keeping it there would be in-case we change the implementation of `assert_fa2_token_balance` in the future, though a comment to this end would probably also suffice.

Issue #284, opened 2021-11-02
Testing different configurations


gkaracha:
The closer we get to addressing #171 the more difficult it becomes to test Checker. Some thoughts:

**Unit tests**: Our testsuite contains many unit tests that tightly check edge cases, which are specific to our current choice of parameters (specified in `checker.yaml`). Any change in the value of these parameters can make the testsuite fail, since the edge values change.

**Property-based tests**: Most property-based randomized tests should be able to pass even with a different choice of parameters, since these tests mostly describe general properties we expect to hold, that are mostly insensitive to the specific choice of parameters. However, now that we generate more code at build time, even property-based tests cannot always be shared. For example, different curves for the drift derivative (see #283) can be expected to have different properties.

**Mutation testing**: Apart from the challenge of keeping test coverage high while the codebase changes so fast, the results of mutation testing also become less useful:
* We now rely on `tezWrapper` which is used meaningfully only in e2e tests. Mutating it thus does not yield meaningful results.
* Soon we'll have a ctez wrapper too (see https://github.com/tezos-checker/checker/issues/279#issuecomment-956115938), so that part will also be tested solely in the e2e tests.
* Several files have turned into jinja templates, which at the moment (reasonably) are not mutated.

Certainly Checker will be shipped with a set of default values and we need the testsuite to have sufficient coverage for the default configuration, but ideally we'd have a way to test a wider range of configurations. It is unclear how to achieve that at the moment, but it looks like we'll need to separate (a) general properties that we expect to hold independently of the choice of values in `checker.yaml` (assuming that they are reasonably set, of course) from (b) configuration-specific tests.

gkaracha:
Now that all three generalizations (#213, #283, #311) are implemented we can start thinking about how to go about improving our testing setup to accommodate all the different possible configurations. Currently (as of 89b06df) we've kept only the tests that can pass unaltered under **_all_** the following configurations (which are the configurations tested on CI):
* `curve_type: bang-bang`, `tracking_type: index`, `collateral_type: tez`
* `curve_type: bang-bang`, `tracking_type: index`, `collateral_type: fa2`
* `curve_type: bang-bang`, `tracking_type: token`, `collateral_type: fa2`

Existing tests whose behavior is not identical for the above configs have been temporarily removed/commented out but they should be the first ones to come back. Lest we forget, let's list them here

[...]

PR #311, opened 2021-12-10
Support for replicating tokens on chain

### LINKED ISSUE #263: Generalize checker to be able to replicate tokens on chain
Taken from #171 (Part 3):

> We can distinguish two big classes of contracts: those that attempt to replicate an index provided by an oracle (index-based), and those that attempt to replicate an existing token on the chain (token-based)
> 1. For index-based we use the index, the ctez / tez cfmm price, and the ctez / kit cfmm price, with the ctez / kit cfmm being subsidized by the checker contract.
> 2. For token-based, the setup is typically a bit different. Suppose we want to replicate USDS using tzBTC as collateral. In that case, we would want to derive prices from 3 cfmms, USDS / cUSDS, tzBTC / ctez and USDS / ctez, with the USDs / cUSDS pair being subsidized by the checker contract.

Though the technical details of how to allow checker at build/deployment time switch from index-based (current design) to token-based mode are currently unclear, I'm opening this issue so that we can keep all relevant discussion in one place.

Though perhaps generalization in this axis is orthogonal to #213, I'd recommend we address this after #213 is completed.

gkaracha:
Closes #263.

* Implement a mock cfmm oracle to use in tests. To ensure that our tests can catch forgotten changes (i.e., avoid mixing it up with the usual mock oracle), the name of the price entrypoint is `%getCfmmPrice` (as opposed to `%getPrice` which is the one we use for the mock oracle). This is not very descriptive and does not indicate the units we expect the fraction to have (tok/ctok), but it's only a placeholder for now. We might want to change this in the future, or even perhaps get it from the yaml file at build time.
  Also: currently both mock oracles (index and cfmm) have an `update` entrypoint but since this is not part of Checker's spec I saw no good reason to deviate on the name of this one too.
* Extend the yaml specification with another field (`tracking_type`) which can take one of two values: `index` or `token`.
* Generalize the type of the (index-)oracle to a pair of `nat`s (CHF in TEZ, (TEZ/CHF)). This reduces the number of differences between token-based and index-based instances quite some.
* Turn the calculation _of the target_ into jinja templates. The different price calculation does manifest itself in the drift adjustment (see https://github.com/tezos-checker/checker/issues/263#issuecomment-969045606), but indirectly, through the target. Hence the target became a jinja template and not the drift derivative. 
* Add `docs/prices.md` with some notes about price calculations, taken from https://github.com/tezos-checker/checker/issues/263#issuecomment-969045606.
* Extend CI to run all the tests with all the three possible configurations: `checker-index-tez.yaml`, `checker-index-fa2.yaml`, and `checker-token-fa2.yaml`.


gkaracha:
OK, I think this one is finally ready for review @dorranh! The description is also now up to date.

After merging this I think we'll be at the point where the README should have separate build/deployment/testing instructions per configuration. Hopefully the API won't really change after this PR.

I had to deactivate three tests that are configuration-specific for CI to pass, but we should be able to renable those tests once #284 is addressed (partially or completely; I might be able to fix that rather easily, being on the OCaml side). Either way, I'd go ahead and merge this PR without them for now.

src/targetCalculation.mli
```diff
@@ -0,0 +1,4 @@
+open Common
+open FixedPoint
+
+val compute_current_target : (* current_q *) fixedpoint -> (* current_index *) fixedpoint -> (* current_kit_in_tok *) ratio -> fixedpoint
```
dorranh, line 4: I actually didn't realize that you could make inline comments in function signatures like this

scripts/builder/checker_builder/templates/tokenTargetCalculation.ml.jinja
```diff
@@ -0,0 +1,18 @@
+(* Automatically generated by checker_builder. DO NOT EDIT DIRECTLY! *)
+
+open Common
+open FixedPoint
+
+(** Calculate the current target based on the current quantity, and the current
+    price of kit in ctok (as given by the CFMM).
+    {[
+      target_{i+1} = FLOOR (q_{i+1} / kit_in_ctok_{i+1})
```
dorranh, line 9: Is the target still dimensionless if we've removed `index` from this equation? To me this reads: `[1/kit] / [ctok/kit] => [1/ctok]`. Perhaps q has different units in the case of a token-based checker? Not sure if this has implications for the system though.

src/checker.ml
```diff
@@ -872,12 +873,14 @@ let entrypoint_touch (state, _: checker * unit) : (LigoOp.operation list * check
...
-let entrypoint_receive_price (state, idx: checker * Ligo.nat) : (LigoOp.operation list * checker) =
+let entrypoint_receive_price (state, oracle_price: checker * (Ligo.nat * Ligo.nat)) : (LigoOp.operation list * checker) =
   assert_checker_invariants state;
   if !Ligo.Tezos.sender <> state.external_contracts.oracle then
     (Ligo.failwith error_UnauthorisedCaller : LigoOp.operation list * checker)
   else
-    let price = fixedpoint_of_ratio_floor (make_ratio (Ligo.int idx) tok_scaling_factor_int) in
+    let price =
+      let (num, den) = oracle_price in
+      fixedpoint_of_ratio_floor (make_ratio (Ligo.int num) (Ligo.int den)) in
```
dorranh, line 883: Maybe we should add a comment here to make it extra obvious that we are losing some precision?

scripts/builder/checker_builder/templates/tokenTargetCalculation.ml.jinja
dorranh: I just re-read the new document covering the units of measure and am still a bit confused about which units `q` is supposed to have. Perhaps we could discuss this offline as well.

dorranh (approved):
Overall looks good to me! I just had one question regarding units of measure. If the comment is not actually an issue then I am happy merging :slightly_smiling_face:

src/checker.ml
gkaracha: Ah yes, I added a small note in a9d6a88 :+1:

scripts/builder/checker_builder/templates/tokenTargetCalculation.ml.jinja
gkaracha: Yeah, let's discuss this more offline. For reference, here's how I've been thinking about the two cases:
* index-tracking CHF
  ```
  index  : tok/CHF
  p_cfmm : tok/kit
  => (index/p_cfmm) : kit/CHF
  ```
* token-tracking tzCHF
  ```
  p_cfmm : tzCHF/kit
  => (1/p_cfmm) : kit/tzCHF
  ```

gkaracha:
@dorranh shall I go ahead and merge this?

dorranh:
@gkaracha, yep I think it's good to go following our separate discussion of the `target` calculation.

PR #314, opened 2021-12-30
Upgrade sandbox to hangzhou

### LINKED ISSUE #297: Update protocol to Hangzhou

dorranh:
Upgraded `flextesa` to a more recent version and our sandbox to use protocol `Hangzhou`. Closes #297.


dorranh:
Closing in favor or #315

PR #315, opened 2021-12-30
Upgrade sandbox protocol to hangzhou

dorranh:
Updated `checker-builder` to allow for code to be generated in a user-specified directory and used this within our Earthly build to eliminate the need for explicitly passing generated module names. After reading about buildkit more my hunch is that this will also fix the sporadic `Error: build target: build main: bkClient.Build: failed to solve: duplicate output 0` errors we get when saving code which seem to primarily occur in the `+generate-code` target (https://github.com/earthly/earthly/issues/1406). The `duplicate output` code seems to be related to buildkit attempting to write outputs under the same path (even if they have different file names).


dorranh:
Very interestingly, it looks like these changes do indeed fix the `duplicate output` Earthly error which is blocking #314. Since this PR is based off of the same branch, and I would prefer not to waste so much CI time I'm going to go ahead and land this PR and close the other one.

gkaracha:
Nice, thanks!

I am quite surprised btw by the [gas cost differences](https://github.com/tezos-checker/checker/pull/315#issuecomment-1003067674); all gas costs dropped (some of them significantly), but those of `checker%touch` went up. Interesting how the new protocol changes the gas cost calculation :)

PR #319, opened 2022-01-18
Use a different curve when tracking_type=token

### LINKED ISSUE #318: Use a flatter curve for the CFMM when tracking_type is set to token
This is the continuation of https://github.com/tezos-checker/checker/issues/263.

When we are tracking a token on chain (i.e., when `tracking_type: token` in `checker.yaml`) it is desirable for the CFMM to use a flatter curve with less slippage, like the one currently used by ctez (branch `next_next`). See [here](https://github.com/tezos-checker/ctez/tree/next_next#curve) and [here](https://github.com/tezos-checker/ctez/blob/3b41b4f374d95563edbff641af030ab929a9a5ae/cfmm_tez_ctez.mligo). The most straightforward way to implement this is probably to have two CFMM implementations (`indexCfmm.ml.jinja` and `tokenCfmm.ml.jinja`, for the CPMM and the flat one, respectively) and copy/generate the appropriate `cfmm.ml` at build time.

gkaracha:
This is currently work in progress, meant to close #318.


dorranh:
I've had a look at the existing changes on this branch and implemented the remaining steps of:

1. Plugging the new "flat curve" into the CFMM logic in `tokenCfmm.ml.jinja`
2. Adding custom error messages for dx / dy estimation functions
3. Disabling a limited set of tight unit tests which do not hold for both cfmm implementations

While I think this PR is in decent shape, I would caution that these changes to Checker's CFMM for cases where Checker is configured to track on-chain tokens are new and still experimental. As such they will need additional rounds of testing and verification in the future.

PR #320, opened 2022-01-21
Update vendored ctez to next_next

dorranh:
So far we've been running our end to end tests, etc. against ctez's `next` branch. This PR updates the vendored version to `next_next`, which is the more recent version of ctez.

Also updated the approach used for fetching the various `tezos-*` binaries in the Earthly build for `flextesa` since the script provided with flextesa is flaky (refers to transient build artifacts which can expire instead of actual releases).

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,4 +1,4 @@
 [submodule "vendor/ctez"]
 	path = vendor/ctez
 	url = https://github.com/tezos-checker/ctez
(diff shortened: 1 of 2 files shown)


dorranh:
Going to go ahead and land this one since CI is passing and we can't actually build the `flextesa` target on the current `master` without the updates in this PR.

PR #321, opened 2022-02-11
Switch ligo to vendored version

dorranh:
Due to issues with configurations within ligo and the large size of Checker's entrypoints, we've had to compile the contract using a patched version of ligo. Thankfully, [recent updates](https://gitlab.com/ligolang/ligo/-/issues/1227#note_820191587) have made it possible to compile checker with upstream ligo. This PR switches over our builds to use upstream ligo instead of our fork.

One particularity of this PR is that since ligo does not currently publish arm binaries, we vendor / rebuild it within our Earthfile in order to continue to support M1 Mac users.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,3 +2,6 @@
 	path = vendor/ctez
 	url = https://github.com/tezos-checker/ctez
 	branch = next_next
+[submodule "vendor/ligo"]
+	path = vendor/ligo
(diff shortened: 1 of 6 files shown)


dorranh:
Going to go ahead and land these changes. Note that we might want to homogenize our logic for pulling in `ligo` and `flextesa` since the latter is pulled within our `Earthfile`, but this is something we can tweak later as needed.

PR #323, opened 2022-04-07
fix tests and readme setup instructions

emturner:
tested on linux
fixes initial setup & the deprecation error when running the tests

closes #322

PR #324, opened 2022-04-07
Update doc workflow

dorranh:
Following some discussions around why the `doc` GitHub Actions workflow was failing for #323, I noticed that the `doc` workflow definition follows a different pattern for authenticating to GitHub and deciding when to push images to the Earthly cache. I'm not sure if this is the only reason why the build in #323 fails, but either way these workflows should be following the same approach. This PR updates the `doc` workflow to use the same methodology as `ci`.


emturner:
closing in favour of #323, which includes these changes

PR #326, opened 2022-08-31
quick jakarta fix

ruquant:
quick fix in order to deploy checker to jakarta testnet


purcell:
Rather than fixing up the generated Ligo, you shoul make the modifications in the `.ml` and other files which produce it. That should be easier overall.

